PR #23, opened 2023-08-30
KIN-110 Login form logic - login page

Mauriciocr207:
[KIN-110](https://producepay.atlassian.net/browse/KIN-110) Medium [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

A user needs a validation process to acces to administration panel in order the user can be logged.

### Solution

Was created a validation class that controls all related to inputs validation and validation of credentials stored on database. Also it was added all responses on the view in order user can know if the validation process was corrected

### Requirements checklist

- [x] The form must validate that the fields aren't empty or invalid, this means, that the credentials obtained from the email and password fields must be complete and consistent. 
- [x] The form must validate that the email exists in a register on the database
- [x] The form must validate that the password is correct.
- [x] If the credentials do not pass the validations, the errors must be sent to the form view.
- [x] If the credentials pass the validations, the user must be redirected to the administration panel.

### Testing

- [x] Run the UserSeeder on terminal using: `php spark db:seed UserSeeder`.
- [x] Write in login form the credentials of any user you created on database with UserSeeder. By default all users have "fake_password" as their password.

## Demo

[KIN-110]: https://loopcrack.atlassian.net/browse/KIN-110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


FernandoJoachin (commented):
In the validation of the existence of the user in the database, you can use the "is_not_unique" validation that codeigniter brings instead of the query builder

app/Validation/LoginValidation.php
```diff
@@ -0,0 +1,41 @@
...
+        ]
+    ];
+
+    protected $invalid_credentials_message = "Email o contraseña inválidos";
+
+    public function validateCredentials($email, $password) {
+        $userModel = new UserModel();
+        $user = $userModel->where("email", $email)->findAll(1)[0] ?? false;
+        if($user) {
+            if($email == $user["email"]) {
```
Cristianpan, line 30: Please check if this condition is really necessary.

app/Validation/LoginValidation.php
```diff
@@ -0,0 +1,41 @@
...
+        "password" => [
+            "required" => "La contraseña es obligatoria",
+        ]
+    ];
+
+    protected $invalid_credentials_message = "Email o contraseña inválidos";
+
+    public function validateCredentials($email, $password) {
+        $userModel = new UserModel();
+        $user = $userModel->where("email", $email)->findAll(1)[0] ?? false;
```
Cristianpan, line 28: In this statement it would be better to use the first() method instead of findAll(). You can find detailed information 
[here](https://codeigniter.com/user_guide/models/model.html#first)

CheloChidoone (changes requested):
It looks good but I agree with the comments Cristian made, check the method you are using to find the user and if you already found the user by email, checking the email again seems redundant.

app/Controllers/CtrlLogin.php
```diff
@@ -21,7 +23,22 @@ public function passwordReset($token) {
...
-        return "Sending access credentials..";
+        $loginValidation = new LoginValidation();
+
+        $data = $this->request->getPost();
+
+        $email = trim($data["email"]);
+        $password = trim($data["password"]);
+
+        if($loginValidation->validateInputs($data)) {
+            if($user = $loginValidation->validateCredentials($email, $password)) {
```
CarlosJCD, line 34: It would not be better to wrap this two conditionals in one if statement?

CheloChidoone: I think it will be easier to read separately like it is now because in the second condition he is also assigning a value to the variable `$user`, and it is only two nested if statements so it isn't a big problem here but if we needed more conditions then we don't want to use nesting if statements too much because it will become complex and difficult to understand

LUGAMAFE (approved):

Mauriciocr207 (commented):
As agreed with the team at this moment, it is decided use Exceptions to handle validation of this login

CheloChidoone (approved):

Cristianpan (approved):

CarlosJCD (approved):

PR #25, opened 2023-08-30
KIN-115 Add change password page

Cristianpan:
[KIN-115](https://producepay.atlassian.net/browse/KIN-115) Medium [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

The web app requires a page for users to enter their new password and reset it

### Solution

The page was created so that the user can enter their new password and reset it. 

### Requirements checklist

- [x] The form must obtain the following data:
       - New password
       - Confirmation of the new password
- [x] It must contain a button to change password

### Testing
Note: Access the address "/login/password/reset/token" to display the page.
- [x] Fill both fields without the passwords matching and verify that the "Las contraseñas deben de coincidir" error message appears.
- [x] Fill both fields correctly and submit the form, check that the message "Change password with token: token..." appears.

## Demo
Video demo
[kinub-cm-change-password-reset-video](https://www.youtube.com/watch?v=SCB_HouXz98)

[KIN-115]: https://loopcrack.atlassian.net/browse/KIN-115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


LUGAMAFE (approved):
Hey @Cristianpan quick suggestion I think we will have to have a base controller for the emails in this case `appControllersCtrlEmail.php` but I am seeing that inside you are using methods like this, it would be better if we can have different controllers for the emails extending from `appControllersCtrlEmail.php` and move this methods to that place
```
    public function sendEmailToResetPassword() {
        return "sending email to reset password...";
    } 
```

LUGAMAFE:
@Cristianpan igual faltan las tags de tu PR

Mauriciocr207 (approved):

CarlosJCD (approved):

FernandoJoachin (approved):

PR #53, opened 2023-09-14
KIN-88 Equipment Layout Public View Front End

CarlosMay7:
[KIN-88](https://loopcrack.atlassian.net/browse/KIN-88) Medium [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

The site needs a place where the users can find all the types of equipment they have

### Solution

Creating a page where all the different categories are listed is a way to give the users the information they need

### Requirements checklist

- [x] It must have a heading associated to the page

- [x] It must have a background video associated with the page.

- [x] Each category must be represented as a cell of a grid. Each cell must contain the following:

    - An icon related to the category
    - The category’s name.

- [x] It must be responsive

### Testing

Describe any tests (manual or automated) executed to verify your changes

- [x] Checked the responsiveness
- [x] Check the animations
- [x] Check the video is playing

## Demo

image
image
image
image
image

[KIN-88]: https://loopcrack.atlassian.net/browse/KIN-88?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CarlosMay7:
I created a placeholder route "equipment" in order to make and try this design. When this feature is accepted I will delete that route

CarlosMay7:
I also think the spacing is kind of weird but while it is getting closer to the max (9 categories I think) it gets better

app/Config/Routes.php
```diff
@@ -41,6 +41,10 @@
 $routes->post('/email/soporte', 'CtrlEmail::sendSupportEmail');
 
 
+
+$routes->get('/equipment', 'CtrlPublicPages::viewEquipment');
```
CarlosJCD, line 45: I am curious, why using a placeholder instead of the route specified on the documentation?

CarlosJCD (approved):
Great job. Be sure to replace the placeholder route with the one specified on [the documentation](https://loopcrack.atlassian.net/wiki/spaces/Kinub/pages/3178510/Routes).

CheloChidoone (commented):
It looks fine but I also wonder why you used a placeholder route instead of the actual route specified?

app/Config/Routes.php
CarlosMay7: Sorry, I wasn't sure if I could create one more function in the controller and put a new route on the router so I put it that way but I'll change it, thanks

CarlosJCD: Don't worry, just asking by mere curiosity. Just change the route to the one on the doc and that would be it. ;)

app/Views/public/equipment.php
```diff
@@ -0,0 +1,65 @@
...
+    <h1 class="equipments__heading">Equipos</h1>
+
+    <div class="equipments__content">
+        <video class="equipments__video" autoplay muted loop> 
+            <!--Add more video formats-->
+            <source src="assets/video/equipment-video.mp4">
+        </video>     
+
+        <div class="equipments__grid">
+
```
Mauriciocr207, line 19: image
I see that the grid is taken width 100%, it could be better if the `class="equipment_grid"` have a max-width. And for the video, it could be better if it extends to all width.
Also, I see you use gap: 2rem and row-gap:2rem, you can pass row and column gap to gap property, like: `gap: 2rem 4rem` with you have `row-gap:2rem` and `column-gap:4rem`

Mauriciocr207 (commented):
image
I see that the grid is taken width 100%, it could be better if the `class="equipment_grid" have a max-width. And for the video, it could be better if it extends to all width.

CheloChidoone (approved):
It looks better, good job!

src/sass/public/UI/equipments/_equipments.scss
```diff
@@ -0,0 +1,169 @@
...
+
+    &__icon,
+    &__name {
+        transition: transform 300ms ease-in-out;
+    }
+
+    transition: backdrop-filter 200ms ease-in-out;
+
+    &:hover {
+        backdrop-filter: blur(50px);
```
LUGAMAFE, line 155:
```suggestion
        backdrop-filter: blur(50px) hue-rotate(15deg) saturate(1.8);
```

add this to match the hover effect in the designs

src/js/public/equipments.js
LUGAMAFE, line 1: @CarlosMay7 hahahaha bro you went too far, delete this file.
The same thing you did here with JS can and should be done with css and mediaqueries, you literally started to mimic the behavior of media queries, do you realize?

src/sass/public/base/_global.scss
```diff
@@ -34,4 +34,10 @@ ul {
     max-width: 70rem;
     width: 90%;
     margin: 0 auto;
+}
+
+[class*="__container"] {
+    width: 95%;
+    max-width: 120rem;
+    margin: 0 auto;
```
LUGAMAFE, line 42: this is not the correct file to add this css

LUGAMAFE (changes requested):
Very good design @CarlosMay7 I would take a little more care of the margins that are not equal to the design but it is only a nitpick not a blocker.

I left you some comments

src/js/public/equipments.js
CarlosMay7: I did it that way thinking that there will be times when there is one left over in the rows so that thing will have to be placed in the middle, but there will also be occasions when there are 2 left over or they are full, so it doesn't apply and I didn't find a way to put those conditions in css so I opted for js, should I continue looking for options to put those conditions in css?

src/sass/public/UI/equipments/_equipments.scss
```diff
@@ -0,0 +1,169 @@
...
+            @include m.grid(2, 4rem 2rem);
+            padding: 7rem 4rem;
+        }
+        @include m.tablet{
+            padding: 6rem 5rem;
+            row-gap: 3rem;
+        }
+        @include m.desktop {
+            padding: 4rem 5rem;
+            @include m.grid(3, 4rem 2rem);
```
LUGAMAFE, line 49:
```suggestion
           @include m.flex-columns(3, 2rem, 4rem);
```

include this new mixin I created. 

don't forget to add this mixin in the corresponding _mixins.scss file

``` scss
@mixin flex-columns(
  $columns,
  $column-gap,
  $row-gap,
  $items-max-height: none,
  $justify: center,
  $align-content: flex-start
) {
  display: flex;
  flex-wrap: wrap;
  justify-content: $justify;
  align-content: flex-start;
  row-gap: $row-gap;
  column-gap: $column-gap;
  > * {
    flex-basis: calc((100% - $column-gap * ($columns - 1)) / $columns);
    max-height: $items-max-height;
  }
}
```

`flex-columns` Mixin

1. **$columns**: Specifies the number of columns.
2. **$column-gap**: Gap between columns.
3. **$row-gap**: Gap between rows.
4. **$items-max-height**: Maximum height for items. It defaults to none.
5. **$justify**: Sets the justify-content property. Defaults to center.
6. **$align-content**: Sets the align-content property. Defaults to flex-start.

It uses Flexbox to create a layout, utilizing the flex-wrap, justify-content, align-content, row-gap, and column-gap properties. It then calculates the flex-basis for child elements to properly place them into columns.

`grid` Mixin

1. **$columns**: Specifies the number of columns.
2. **$gap**: Gap between grid items.

It uses CSS Grid to create a layout using grid-template-columns and gap properties. It's much more straightforward and relies on the built-in Grid layout capabilities to place items into columns.

**Differences**
**Flexibility**: The `flex-columns` mixin provides more control over how items should be laid out. You can control justification, content alignment, and even the maximum height of items.

**Code Complexity**: _grid_ is simpler and has less code, making it easier to understand. However, it's also less flexible.

**Cross-Browser Support**: Flexbox has better support in older browsers compared to CSS Grid. Although as of my last update in September 2021, this is becoming less of a concern.

**Responsibility**: In `flex-columns`, you have to calculate the width of the columns yourself (using `flex-basis`). In `grid`, the grid layout takes care of that for you.

**Which is better?**
"Better" can be subjective, depending on what you need:

For simplicity and modern layouts: Use the grid mixin. It's cleaner and leverages newer CSS standards.

For more control and possibly better browser support: Use the flex-columns mixin.

src/js/public/equipments.js
LUGAMAFE, line 1: now you can delete this file and use `flex-columns` mixin instead

app/Views/public/equipment.php
```diff
@@ -0,0 +1,74 @@
+<?php $this->extend('public/templates/layout'); ?>
+
+<?php $this->section('css'); ?>
+<link rel="stylesheet" href="assets/css/public.min.css" type="text/css">
+<?php $this->endSection(); ?>
+
+<?php $this->section('js'); ?>
+<script src="assets/js/public/equipments.js"></script>
+<?php $this->endSection(); ?>
```
LUGAMAFE, line 9:
```suggestion

```

delete if no needed for the moment

LUGAMAFE (changes requested):
I created a new mixin that will help you to remove the js file @CarlosMay7  and I also added some changes to improve the sass and match the design better.

FernandoJoachin (approved):
Good job!!

LUGAMAFE (approved):
Nice Job @CarlosMay7

Oswaldo-Chan (approved):

PR #60, opened 2023-09-20
KIN-218 Add routing access validation in admin and login section

Cristianpan:
[KIN-218](https://loopcrack.atlassian.net/browse/KIN-218) Small [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

With an admin panel with limited access, it is needed a mechanism that allows to validate the access to certain routes

### Solution

Filters have been created to validate the user's session and access to certain paths depending on the type of user logged into the administration panel.

### Requirements checklist

- [x] It must validate that the user has the correct access to the route. 
- [x] The user must be redirected to the login page if he tries to access a private view without being logged in. If the user then logs in, he should be redirected to the route that he were trying to access from the beginning.
- [x] The user must be redirected to the home page of the admin panel if it has already been logged in and tries to access a route of the admin account.
- [x] If the user is already logged and tries to access to the login page, it must be redirected to the home page of the admin panel.

### Testing

- [x] Access any route above "/admin/*" without being logged in and the application should redirect you to "/login". If you then log in, you should be redirected to the route you were trying to access.  
- [x] Log in with an account and you should be able to access the administration panel and the associated routes.
- [x] If you are already logged in, go to the "/login" route and you should be redirected to the start of the administration panel.
- [x] Log in with a "basic" administrator account and try to access the path "/admin/users", the app should redirect you to the start of the administration panel.
Note: In order to test this last point, you can add the following in "app/config/Routes.php": 
`$routes->get('/admin/usuarios', 'CtrlCategory::viewCategories'); 
`

## Demo

[KIN-218]: https://loopcrack.atlassian.net/browse/KIN-218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


FernandoJoachin (commented):
When I tried to enter the dashboard without having logged in I got this error. This shouldn't happen, right?

[](https://postimg.cc/23F2c4j0)

CheloChidoone (approved):
Good job Cristian!

LUGAMAFE (changes requested):
Hey @Cristianpan  I love this new filters utility that codeigniter has to handle access to routes.
I just feel there is one case that you are not keeping in mind.

We need to redirect a user trying to access a protected path to the login page, YES! But once he logs in, we have to redirect him to the path he wanted to access in the first place.

For example if he tries to access the url:` http://localhost:8080/admin/soluciones`
but he is not logged in yet 
we take it to the url:
`http://localhost:8080/login`
up to here everything correct. 
But now after logging in we are redirecting him to the admin panel:
`http://localhost:8080/admin`
when we should be taking it to the route that he was trying to access from the beginning, in this case:
`http://localhost:8080/admin/soluciones`

you could to something similiar to this:
```php
class AuthFilter implements FilterInterface
{
    public function before(RequestInterface $request, $arguments = null)
    {
        $user = session()->get("user"); 

        if (!$user){
            // Guarda la URL actual en la sesión
            $session->set('redirect_url', current_url());
            
            return redirect()->to("login"); 
        }
    }
```

and when the login is finished you could do something similar to this:
```php
public function processLogin()
{
    // Lógica de inicio de sesión

    $session = session();
    
    if ($session->has('redirect_url')) {
        $redirectUrl = $session->get('redirect_url');
        $session->remove('redirect_url');
        return redirect()->to($redirectUrl);
    }
    
    // Si no hay URL en la sesión, redirige a la página por defecto
    return redirect()->to('/admin');
}
```
Please help me by updating the ticket and ticket description with this extra case please. 🙏 Muchas gracias!!

CarlosJCD (approved):
Everything works fine, nice job!

LUGAMAFE (approved):
Nice!!

Mauriciocr207 (approved):

PR #65, opened 2023-09-23
KIN-237 Refactor navigation bar implementation

CarlosMay7:
[KIN-237](https://loopcrack.atlassian.net/browse/KIN-237) Medium [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

The navigation bar implementation should be a block that does not change depending on the device size

### Solution

Change the HTML implementation so instead of being different blocks there is now only one that changes its styles dynamically

### Requirements checklist

- [x] The navigation bar must have the same functionality 
- [x] The navigation bar HTML implementation must be one block
- [x] The link routing must redirect to the respective route
- [x] There must not be a separation between the navigation bar and the "About Us" section

### Testing
- [x] Manually checked the navigation bar behavior through different events like opening, resizing and redirecting
- [x] Manually checked all the routes

## Demo
Demo video
https://github.com/loopcrack-org/Kinub/assets/97546829/eac60b4d-db1c-41b8-a1df-8a16b06f09e9

image
image

[KIN-237]: https://loopcrack.atlassian.net/browse/KIN-237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/js/public/menuMobile.js
```diff
@@ -1,28 +1,48 @@
...
+const displayWidth = document.body.clientWidth;
+
+if (displayWidth<751) {
+    header.classList.add("header-mobile");
+    navigation.classList.add("no-visible");
+}
 
 if(btnMobileMenu){
+    const displayWidth = document.body.clientWidth;
+    if (displayWidth<751) {
```
CarlosJCD, line 15: you could extract this to a function, something like "isMobileView".  Especially considering that you use this expression in more than once.

src/js/public/menuMobile.js
```diff
@@ -1,28 +1,48 @@
...
+        btnMobileMenu.classList.remove("display-none");
+    }
     btnMobileMenu.addEventListener("click", function(){
-        headerMobile.classList.toggle("no-shadow");
+        header.classList.toggle("no-shadow");
         btnMobileMenu.classList.toggle("is-active");
-        if(sidebar.classList.contains("show")) {
-            sidebar.classList.add("hide");
+        if(headerBackground.classList.contains("show")) {
```
CarlosJCD, line 21: you could refactor this event listener to its own function.

src/js/public/menuMobile.js
```diff
@@ -1,28 +1,48 @@
...
+                navigation.classList.add("no-visible");
+                navigation.classList.remove("hide-opacity");
             }, 500);
+
         } else {
-            sidebar.classList.add("show");
+            headerBackground.classList.add("show");   
+            navigation.classList.remove("no-visible");
+            navigation.classList.remove("hide");       
+            navigation.classList.add("show");
```
CarlosJCD, line 28: this four lines of this else statement could be its own function.

src/js/public/menuMobile.js
CarlosJCD, line 21: you could change this to a function with a more intention revealing name

CarlosJCD (commented):
Nice job as always, however, even if it works everything fine I think it is necessary to make an effort to generate a more maintainable and cleaner code. 

I left some suggestions on this, basically refactor to smaller functions with intention revealing names that helps the code readability. However there are a lot of this opportunity areas to improve

CarlosJCD (commented):
Also a found a resolution where the navbar is not visible. I do not know if it is just in my computer, please check it out.
Screenshot 2023-09-23 at 21 49 27

CarlosJCD (approved):
excellent work, it is a lot more readable now!

FernandoJoachin (commented):
Have you already solved the navigation problem?
Because what happens to me is that at a resolution lower than 767px the navigation is not visible.

[](https://postimg.cc/s1YCTXdR)

CheloChidoone (changes requested):
Good Job @CarlosMay7, I didn't see the error @CarlosJCD and @FernandoJoachin mentioned about the navbar not being visible, for me it was visible in those resolutions, I tested in Chrome, Edge, and Firefox. I left some comments so you can improve the organization and simplify some parts of your .js

Cristianpan:
@CheloChidoone , Yes there is a small bug with the navigation bar. I'v attached a demo video of the bug.
[navbar bug](https://youtu.be/hW53pP0f3Y4).

I've been checking the mobileMenu.js script and found the "error". In this function I consider that it wouldn't be entirely correct to use  [`document.body.clientWidth`](https://developer.mozilla.org/en-US/docs/Web/API/Element/clientWidth)
, perhaps it would be better to use [`window.innerWidth`](https://developer.mozilla.org/en-US/docs/Web/API/Window/innerWidth).

```js 
function isMobile(){
    const displayWidth = document.body.clientWidth;
    return displayWidth<768;
}
```

Cristianpan:
> Good Job @CarlosMay7, I didn't see the error @CarlosJCD and @FernandoJoachin mentioned about the navbar not being visible, for me it was visible in those resolutions, I tested in Chrome, Edge, and Firefox. I left some comments so you can improve the organization and simplify some parts of your .js

@CheloChidoone , Yes there is a small bug with the navigation bar. I'v attached a demo video of the bug.
[navbar bug](https://youtu.be/hW53pP0f3Y4).

I've been checking the mobileMenu.js script and found the "error". In this function I consider that it wouldn't be entirely correct to use  [`document.body.clientWidth`](https://developer.mozilla.org/en-US/docs/Web/API/Element/clientWidth)
, perhaps it would be better to use [`window.innerWidth`](https://developer.mozilla.org/en-US/docs/Web/API/Window/innerWidth).

```js 
function isMobile(){
    const displayWidth = document.body.clientWidth;
    return displayWidth<768;
}
```

CheloChidoone (approved):

Oswaldo-Chan (approved):

src/public/sass/UI/header/_header.scss
```diff
@@ -1,176 +1,264 @@
...
+  box-shadow: 0px 4px 15px 0px v.$gray;
+  transition: box-shadow 500ms ease-in-out;
+
+  &__logo {
+    height: 5rem;
+    width: 7rem;
+  }
+
+  &__background {
+    opacity: 0;
```
Oswaldo-Chan, line 24:
```suggestion
    visibility: hidden;
    opacity: 0;
```

src/public/sass/UI/header/_header.scss
```diff
@@ -1,176 +1,264 @@
...
+      &::after {
+        top: 0;
+        opacity: 0;
+        transition:
+          top 200ms cubic-bezier(0.33333, 0, 0.66667, 0.33333),
+          opacity 100ms 220ms linear;
+      }
+    }
+    .header__background {
+      opacity: 1;
```
Oswaldo-Chan, line 61:
```suggestion
      visibility: visible;
      opacity: 1;
```

Oswaldo-Chan (changes requested):
Everything seems to be fine except for one thing, when I visit the page on a computer, it doesn't allow me to click on the elements of the page. This is because the `header__background` is overlaid and always visible.

src/public/sass/UI/header/_header.scss
LUGAMAFE: yeap @Oswaldo-Chan thanks for the correction done!

FernandoJoachin (approved):

Oswaldo-Chan (approved):

PR #82, opened 2023-10-07
KIN-266  Users page layout users page

FernandoJoachin:
### Problem

The admin panel administrator needs to be able to create and delete users who can access the admin panel

### Solution

Create a page that displays a table with all existing users and the actions required for the administrator.

### Requirements checklist

- [ ] It must display the users records as a table, this table must show the following info for each record: Full name of the user, User Email and Delete button
- [ ] The page must contain a register button for new users, this button must be above the users records.

## Demo
[](https://postimg.cc/JHytKn0N)


Mauriciocr207 (approved):

app/Views/admin/users/Users.php
```diff
@@ -0,0 +1,94 @@
...
+                                            </div>
+                                        </td>
+                                    </tr>
+                                <?php } ?>
+                            </tbody>
+                            <tfoot>
+                                <tr>
+                                    <th>ID</th>
+                                    <th>Nombre</th>
+                                    <th>Icono</th>
```
CheloChidoone, line 65: Here it shouldn't be `Icono`, it should be `Email`

CheloChidoone (changes requested):
@FernandoJoachin I think it looks fine but I left a comment for a little mistake you left

app/Views/admin/users/Users.php
```diff
@@ -0,0 +1,94 @@
...
+        <?php echo view('partials/page-title', array('title' => 'Usuarios')); ?>
+
+        <div class="row">
+            <div class="col-lg-12">
+                <div class="card">
+                    <div class="card-header d-flex justify-content-between align-items-center flex-wrap">
+                        <h5 class="card-title mb-2 mb-md-0">Tabla de Usuarios</h5>
+                        <div class="mt-2 mt-md-0 float-md-right">
+                            <a href="/admin/usuarios/crear" class="btn btn-success btn-label waves-effect waves-light rounded-pill">
+                                <i class="ri-add-circle-fill label-icon align-middle rounded-pill fs-16 me-2"></i>Agregar Usuarios
```
CheloChidoone, line 27: Also here it should be `Agregar Usuario` in singular

app/Views/admin/users/Users.php
```diff
@@ -0,0 +1,94 @@
...
+                                        <td>
+                                            <div class="d-flex align-items-center fw-medium">
+                                                <p><?= $user["userFirstName"] . $user["userLastName"];  ?></p>
+                                            </div>
+                                        </td>
+                                        <td><?= $user["userEmail"]; ?></td>
+                                        <td>
+                                            <div class="d-flex justify-content-center">
+                                                <a href="#" class="btn btn-danger btn-icon waves-effect waves-light" data-bs-toggle="modal" data-bs-target="#deleteCategoriesModal" style="width: 100%;"><i class="ri-delete-bin-5-line ri-lg"></i></a>
+                                            </div>
```
Cristianpan, line 56: You should change this code to this, for example: 

```php 
  <td>
                                            <div class="d-flex align-items-center fw-medium">
                                                <p class="text-wrap"><?= $user["userFirstName"] . $user["userLastName"];  ?></p>
                                            </div>
                                        </td>
                                        <td>
                                            <div class="d-flex align-items-center fw-medium">
                                                <p class="text-wrap"><?= $user["userEmail"] ?></p>
                                            </div>
                                        </td>
```

Cristianpan (changes requested):
All good, however I have seen some "flaws" in the galaxy fold view. For example, when displaying the information an overflow occurs in the table and if for example the text name of the name or email is large, which I doubt it is, there is also an overflow. I leave in the comments how you could solve it. 

[Error in galaxi fold](https://postimg.cc/JyYPccQs)

[Error in galaxi fold](https://postimg.cc/SnDrk73g)

CheloChidoone (approved):

LUGAMAFE (changes requested):
Where is the edit button for the users @FernandoJoachin? what if you want to modify the name or the email of the user? we need to think about that

CarlosMay7 (approved):

Cristianpan (approved):

PR #85, opened 2023-10-08
Database implementation Fixes and password column modification

CarlosJCD:
### Problem

It was needed to modify the database for the create user logic, as well to make some few minor fixes on database files

### Solution

Change the user password column to allow null values, as well to make the needed changes on the UserSeeder and MeasurementSolutions class model.

### Requirements checklist

- [ ] The userPassword column of the users table must allow null values
- [ ] The measurement solutions class model file name must match with the class model name
- [ ] The categoryTags seeder must generate categoryId values matching the categories generated by the category seeder.

### Testing

- [ ] Via tableplus or any DBMS, insert a user touple with the password field on null.
- [ ] Validate the category tags seeder by using first the category seeder and then the user tags seeder.

### DIFF
--- a/backup/Database.sql
+++ b/backup/Database.sql
@@ -62,7 +62,7 @@ CREATE TABLE IF NOT EXISTS `Kinub`.`users` (
   `userFirstName` VARCHAR(60) NOT NULL,
   `userLastName` VARCHAR(60) NOT NULL,
   `userEmail` VARCHAR(40) NOT NULL,
-  `userPassword` VARCHAR(60) NOT NULL,
+  `userPassword` VARCHAR(60),
   `userToken` VARCHAR(13) NULL DEFAULT NULL,
   `confirmed` TINYINT NULL DEFAULT NULL,
   `isAdmin` TINYINT NULL DEFAULT NULL,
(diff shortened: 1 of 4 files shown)


CarlosMay7 (approved):

backup/Database.sql
```diff
@@ -62,7 +62,7 @@ CREATE TABLE IF NOT EXISTS `Kinub`.`users` (
   `userFirstName` VARCHAR(60) NOT NULL,
   `userLastName` VARCHAR(60) NOT NULL,
   `userEmail` VARCHAR(40) NOT NULL,
-  `userPassword` VARCHAR(60) NOT NULL,
+  `userPassword` VARCHAR(60),
```
CheloChidoone, line 65: I think it should have `DEFAULT NULL` since it will always start as `NULL` for any new user created, this way we don't have to specify it as `NULL` when creating a new user as it would be always `NULL` by default

CheloChidoone (changes requested):
It looks good @CarlosJCD but I left a comment about the `userPassword` field, see if you agree with me about this change

LUGAMAFE (approved):

CheloChidoone (approved):

Oswaldo-Chan (approved):

PR #90, opened 2023-10-17
KIN-276 adapt login and reset password logic to confirmed accounts

FernandoJoachin:
[KIN-276](https://loopcrack.atlassian.net/browse/KIN-276) Small [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

With the user creation logic defined, we need to use a confirmed account mechanism, so we can have users set their own passwords instead of the admin user.

### Solution

Modify the login logic to check if the user account is confirmed, as well as modify the password reset logic so that the account is confirmed when the new password is set and the account has not been confirmed before.

### Requirements checklist

- [x] Validate when accounts are not confirmed

### Testing

- [x] Sign in with an unconfirmed user
- [x] Reset password with an unconfirmed user

[KIN-276]: https://loopcrack.atlassian.net/browse/KIN-276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/Validation/PasswordEmailValidation.php
```diff
@@ -36,4 +36,14 @@ public function isNotSuperAdmin($admin)
 
         return true;
     }
+
+    public function validateConfirmedAccount($user)
+    {
+        if(!$user || $user["confirmed"] == 0) {
+            throw new Exception("Tu cuenta no ha terminado de ser confirmada. Por favor, revisa tu correo electrónico para finalizar el proceso de confirmación de tu cuenta.");
```
CheloChidoone, line 43: I think the message here should be the same as the login one, it sounds better `Tu cuenta no ha sido confirmada` instead of `Tu cuenta no ha terminado de ser confirmada` because it makes me think I already started the confirmation process when I really haven't done the confirmation.
```suggestion
            throw new Exception("Tu cuenta no ha sido confirmada. Por favor, revisa tu correo electrónico para finalizar el proceso de confirmación de tu cuenta.");
```

CheloChidoone (changes requested):
Good job @FernandoJoachin, it looks like the logic works correctly, I added a little comment about the error messages.

CarlosJCD (commented):
Excellent work, but please keep your PR description clean. The `Resources` and `Demo` section have the values of the PR description template. Please put its correct content, or remove those sections if there is no need of them.

CheloChidoone (approved):

CarlosMay7 (approved):

CarlosJCD (approved):
Thanks for the changes on the pr description. Nice Work

Oswaldo-Chan (approved):

CarlosJCD (approved):

PR #93, opened 2023-10-20
KIN-283 refactoring of the mails sent from the public view forms

FernandoJoachin:
[KIN-283](https://loopcrack.atlassian.net/browse/KIN-283) Small [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

The technical support and contact email was refactored

### Problem

Emails sent from the contact form and support form in public view do not follow the established format standard and cause difficulties in saving them in the database.

### Solution

Refactor public view contact and support form emails so that the html content can match the standard and be saved to the database.

### Requirements checklist

- [x] Follow the established standard to create emails
- [x] The styles must be in the html tags to be able to store the email in the database

### Testing

- [x] Complete the technical support form and check that the email you received is correct
- [x] Complete the contact form and check that the email you received is correct

## Demo

[](https://postimg.cc/0rC5mWRZ)

[](https://postimg.cc/gwRm04HV)

[KIN-283]: https://loopcrack.atlassian.net/browse/KIN-283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CheloChidoone (commented):
@FernandoJoachin I don't know if using an inline style is the best way to do this.
@LUGAMAFE do you think the inline style is correct to use for the emails or could we have the CSS apart from the HTML?

FernandoJoachin:
> @FernandoJoachin I don't know if using an inline style is the best way to do this. @LUGAMAFE do you think the inline style is correct to use for the emails or could we have the CSS apart from the HTML?

FernandoJoachin:
> @FernandoJoachin I don't know if using an inline style is the best way to do this. @LUGAMAFE do you think the inline style is correct to use for the emails or could we have the CSS apart from the HTML?

You need to do this in order to save the email along with the styles in the database.

CarlosMay7 (approved):

Mauriciocr207 (approved):

CarlosJCD (changes requested):
Using the inline styles for each tag it would generate problems in the future if you want to maintain or change some style of the email. Why not using a <style> tag instead?

LUGAMAFE (changes requested):
As  @CarlosJCD mentions in this comment https://github.com/loopcrack-org/Kinub/pull/93#pullrequestreview-1702139715 it is a better idea to move the inline css into a inner <style/> tag to make the code maintainable.

CarlosJCD (approved):

LUGAMAFE (approved):

PR #96, opened 2023-10-24
KIN-268 add delete user functionality

FernandoJoachin:
[KIN-268](https://loopcrack.atlassian.net/browse/KIN-268) Small [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

As an administrator of the admin panel, I need to be able to delete the records of the users so that I can restrict the access to those who i decide to delete. 

### Solution

Add functionality in controllers to delete users.

### Requirements checklist

- [x] It must delete the register from the database
- [x] It must return a response to the view indicating if it was successfully deleted or if it was an error on deleting the user.

### Testing

Describe any tests (manual or automated) executed to verify your changes

- [x] Delete a user.  This action should be reflected in the database and in the view.

[KIN-268]: https://loopcrack.atlassian.net/browse/KIN-268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CheloChidoone (approved):
It works well @FernandoJoachin, please remember to add everyone as reviewers for your PR because it looks like you forgot that this time but good job.

CarlosMay7 (approved):
Good Job! Just check the testing message, I think you repeated the first part of it

CarlosJCD (changes requested):
There is a couple of issues with your current PR:

- On a mac device, the confirm alert when deleted an account or either it is not displayed or either it is displayed as an error even if the user was actually deleted.
- We should not be able to delete the account we are logged into, and furthermore an administrator account should not be able to be deleted.

CarlosJCD (approved):
Excellent work!

app/Controllers/CtrlUser.php
```diff
@@ -31,7 +31,16 @@ public function createUser()
 
     public function deleteUser()
     {
-        $isDeleted = true;
+        $POST      = $this->request->getPost();
+        $userId    = $POST['elementId'];
+        $userModel = new UserModel();
+        $user      = $userModel->find($userId);
+        if (! empty($user)) {
```
CheloChidoone, line 38: @FernandoJoachin Check that the user is not the super admin when deleting, even if the superadmin isn't on the users table the validation should also be done here just in case.

CC: @Mauriciocr207

app/Views/admin/users/Users.php
```diff
@@ -47,25 +47,27 @@
...
-                                            </div>
-                                        </td>
-                                        <td>
-                                            <div class="d-flex justify-content-between">
-                                                <a href="#" class="btn btn-primary btn-icon waves-effect waves-light" style="width: 48%;"><i class="ri-edit-2-fill ri-lg"></i></a>
-                                                <a href="#" class="btn btn-danger btn-icon waves-effect waves-light" data-bs-toggle="modal" data-bs-target="#deleteUsersModal" style="width: 48%;"><i class="ri-delete-bin-5-line ri-lg"></i></a>
-                                            </div>
-                                        </td>
-                                    </tr>
+                                    <?php if ($user['confirmed'] !== '1') { ?>
```
CheloChidoone, line 50: @FernandoJoachin This is wrong, I think you could filter out the superadmin user from the consultation instead of here on the view, but this is wrong anyway confirmed is not the correct field to filter out the superadmin.

CC: @Mauriciocr207

CheloChidoone (changes requested):
@FernandoJoachin We found some changes for you to check

CC: @Mauriciocr207

app/Controllers/CtrlUser.php
Mauriciocr207: Also, in your app/Controllers/CtrlUser.php you have
 ```php
public function viewUsers()
    {
        $userModel = new UserModel();
        $users     = $userModel->findAll();

        return view('admin/users/Users', ['users' => $users]);
    }
```
But you can modify the consult to not show the admin user, something like this
 ```php
public function viewUsers()
    {
        $userModel = new UserModel();
        $users     = $userModel->where("isAdmin", "0")->findAll();

        return view('admin/users/Users', ['users' => $users]);
    }
```

src/admin/js/deleteElement.js
```diff
@@ -0,0 +1,12 @@
+function setIdInModal() {
+  const deleteButtons = document.querySelectorAll('[data-id]');
+  deleteButtons.forEach(function (deleteButton) {
+    deleteButton.addEventListener('click', function () {
+      const elementId = this.getAttribute('data-id');
+      const deleteButtonModal = document.getElementById('elementId');
```
Mauriciocr207, line 6: Maybe here it´s better use 
```javascript
document.querySelector("#elementId");
```
querySelector is more remcomended for this situation.

src/admin/js/deleteElement.js
```diff
@@ -0,0 +1,12 @@
...
+      console.log(deleteButtonModal);
```
Mauriciocr207, line 7: Don´t forget the `console.log`

Mauriciocr207 (approved):

CheloChidoone (approved):

src/admin/js/deleteElement.js
LUGAMAFE: No difference that is opinated recomendation

src/admin/js/deleteElement.js
LUGAMAFE: Yes please @FernandoJoachin delete this

LUGAMAFE (approved):

CarlosMay7 (approved):

LUGAMAFE:
@FernandoJoachin can we close this PR please

CarlosMay7 (approved):

PR #97, opened 2023-10-25
KIN-267 do the create user functionality

FernandoJoachin:
[KIN-267](https://loopcrack.atlassian.net/browse/KIN-267) Medium [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

As an admin panel administrator, I want to create a new user so that they have access to the admin panel.

### Solution

Create a new user creation form to register new users.

### Requirements checklist

- [x] It must validate the form inputs to not be empty.
- [x] It must validate that the email has not been previously registered.
- [x] It must return to the view a response indicating success in signing up a new user or, in case of being errors, it must return the errors.
- [x] it must set a token for the reset password functionality.
- [x] it must set null the user password.
- [x] It must send an email to the user email account notifying him that it's account was created and it is needed to set a password.

### Testing
It is necessary to have a mailtrap account created and replace the .env values with your account credentials.
**Note**: In the mailTpye put html

[](https://postimg.cc/XXsTcHWg)

[](https://postimg.cc/Mvy1J4G7)

- [x] Create a user. When creating the user it should show a success or error alert
- [x] Enter an email that has already been registered in the database.
- [x] Check that the created user cannot log in unless confirmed.
- [x] Check that the email was sent correctly.

[KIN-267]: https://loopcrack.atlassian.net/browse/KIN-267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CarlosMay7 (changes requested):
When I create an account and I have not assigned the password I must not be able to reset it but I can do it and I receive the mail as well, but when I click the set password link an error shows up

CarlosJCD (changes requested):
Nice work!, however we detected an small bug when testing this feature:

- After loggin in with the admin account on the admin panel, if you directly go to create user and you create a valid user, Even if the emal to confirm password is sent and the user record is registered on the database, it does not shows the response confirming success and it does not occurs the redirection to the user page. This only happens with the first user created when loggin in.

FernandoJoachin:
> Nice work!, however we detected an small bug when testing this feature:
> 
> * After loggin in with the admin account on the admin panel, if you directly go to create user and you create a valid user, Even if the emal to confirm password is sent and the user record is registered on the database, it does not shows the response confirming success and it does not occurs the redirection to the user page. This only happens with the first user created when loggin in.

You're sure? Do you run the run dev command previously?

CarlosJCD (approved):

CarlosMay7 (approved):

app/Controllers/CtrlUser.php
```diff
@@ -19,14 +23,118 @@ public function viewUserCreate()
...
+            if (empty($errors)) {
+                $response = [
+                    'title'   => '¡Oops!',
+                    'message' => $th->getMessage(),
+                    'type'    => 'error',
+                ];
+            } else {
+                return redirect()->back()->withInput()->with('errors', $errors);
+            }
+        }
```
Mauriciocr207, line 135: I see that the edit user functionality can´t be reviewed because there is not a view for that functionality. Also, I think this feature is not part of this pull request.

CC: @CheloChidoone

CheloChidoone: @FernandoJoachin You shouldn't mix up what you are doing for each ticket, this PR should only have the create functionality and the other one KIN-285 should have only the edit functionality but you have everything in both.

app/Controllers/CtrlUser.php
```diff
@@ -19,14 +23,118 @@ public function viewUserCreate()
...
+
+            if (! $isSend) {
+                $userModel->where('userEmail', $POST['userEmail'])->delete();
+
+                throw new Exception('Algo salio mal al crear el usuario. Por favor, inténtalo de nuevo.');
+            }
+
+            $response = [
+                'title'   => 'Usuario creado con éxito',
+                'message' => 'La cuenta se ha creado con éxito. Por favor, revise su correo electrónico para establecer su contraseña.',
```
Cristianpan, line 64: The response to this message isn't correct. This should be something like "Por favor notifique al usuario para que revise su bandeja de entrada y confirme su cuenta"

Cristianpan (changes requested):

LUGAMAFE (changes requested):
Same comment as @Mauriciocr207 and @CheloChidoone https://github.com/loopcrack-org/Kinub/pull/97#discussion_r1375342913 we should avoid combine other tickets

app/Controllers/CtrlUser.php
```diff
@@ -19,14 +23,118 @@ public function viewUserCreate()
...
+        try {
+            if (! $validateUser->validateInputs($POST)) {
+                throw new Exception();
+            }
+
+            $userModel = new UserModel();
+            $user      = $userModel->where('userEmail', $POST['userEmail'])->first();
+            $validateUser->existUserEmail($user);
+
+            $token             = uniqid();
```
Cristianpan, line 48: For this, create a Util class like TokerGegerator and create a method that does this functionality.

Cristianpan (changes requested):

CheloChidoone (approved):
Thanks for the changes @FernandoJoachin, good job!

LUGAMAFE (approved):
Ready! thanks @FernandoJoachin

LUGAMAFE:
@FernandoJoachin can we close this PR please

app/Views/admin/users/UserCreate.php
```diff
@@ -90,4 +106,11 @@
     <!-- container-fluid -->
 </div>
 <!-- End Page-content -->
+<?= $this->endSection() ?>
+
+<?= $this->section('js') ?>
+<!-- particles js -->
+<script src="/assets/admin/js/particles.min.js"></script>
+<!-- particles app js -->
+<script src="/assets/admin/js/particles.app.min.js"></script>
```
LUGAMAFE, line 115:
```suggestion

```
Hey @FernandoJoachin me and @Cristianpan noticed that we dont need these scripts please remove them

LUGAMAFE (commented):
Please correct this before merge the branch

Cristianpan (approved):

PR #98, opened 2023-10-27
KIN-285 edit user functionality user page

FernandoJoachin:
[KIN-285](https://loopcrack.atlassian.net/browse/KIN-285) Small [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

As the administrator of the admin panel, I want to persist the changes made to the user credentials on the edit user view so that the changes made are stored on the database without the need of creating a new user and deleting the existing record.

### Solution

Adding the ability to modify database values to the Edit User form

### Requirements checklist

- [x] It must validate that the given info is correct. a.The form must not be empty. b. It must validate that the email has the correct format.
- [x] If the email was changed, it must validate that the new email it is not registered on the database.
- [x] If it passes all validations, it must update the record on the database. a.if the email was changed, it the account must be unconfirmed and it must be sent a confirm email to the new email address.
- [x] It must return a response to the view, the response will contain a message according to the following case: If the record was updated and the email of the user account wasn't changed, the message must indicate success on updating the user record. 
- [x] It must return a response to the view, the response will contain a message according to the following case: If the record was updated and the email of the user account was changed, the message must indicate success on updating the user record and that it is necessary to confirm again the account. 
- [x] It must return a response to the view, the response will contain a message according to the following case: If the new email is already registered on another account, it must return a response indicating that the new email address is already registered and because of that it is not possible to persist the changes. 
- [x] It must return a response to the view, the response will contain a message according to the following case: If it occurred an error on the process, the message must indicate that it occurred an error and to try again later

### Testing

- [x] Edit only first or last name fields
- [x] Edit the email field but with one already registered
- [x] Edit the email field with one that has not been registered

[KIN-285]: https://loopcrack.atlassian.net/browse/KIN-285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/Controllers/CtrlUser.php
```diff
@@ -29,6 +41,62 @@ public function createUser()
...
+
+                $isSend = EmailSender::sendEmail('Kinub', 'kinub@gmail.com', $POST['userEmail'], 'Cuenta Creada de Kinub', 'templates/emails/createUserAccount', $POST);
+
+                if (! $isSend) {
+                    throw new Exception('Algo salio mal al editar el usuario. Por favor, inténtalo de nuevo.');
+                }
+
+                $response = [
+                    'title'   => 'Edición exitosa',
+                    'message' => 'La cuenta se ha actualizado con éxito. Por favor, revise su correo electrónico para establecer su contraseña.',
```
CheloChidoone, line 78: @FernandoJoachin Change message to this like on the create functionality `"Por favor notifique al usuario para que revise su bandeja de entrada y confirme su cuenta"`.

app/Controllers/CtrlUser.php
```diff
@@ -29,6 +41,62 @@ public function createUser()
...
+                $response = [
+                    'title'   => 'Edición exitosa',
+                    'message' => 'Se ha editado el usuario correctamente',
+                    'type'    => 'success',
+                ];
+            } else {
+                $user = $userModel->where('userEmail', $POST['userEmail'])->first();
+                $validateUser->existUserEmail($user);
+                $token                = uniqid();
+                $POST['userToken']    = $token;
```
CheloChidoone, line 66: @FernandoJoachin Please use the token generator util class like you did in the create functionality too.

CC: @CarlosMay7

CheloChidoone (changes requested):
@FernandoJoachin We left you a pair of comments to check.

CC: @CarlosMay7

LUGAMAFE (changes requested):
@FernandoJoachin this Pr should not be open as we need to finish merge this https://github.com/loopcrack-org/Kinub/pull/97 first, also as you can notice this PR is not updated with the main branch. An it has conflicts.

CarlosMay7 (approved):

CheloChidoone (approved):
@FernandoJoachin Good job, it works well, thanks for making the changes.

Oswaldo-Chan (approved):

CarlosJCD (approved):
Nice Work!

Mauriciocr207 (approved):

LUGAMAFE (approved):
image
Hey @FernandoJoachin I found a problem related to the user table logic, a user who just confirmed his account and access to the admin panel and then to the user view, can delete himself as well as edit his own information. 
This should not be allowed. C
ould you please solve this problem in the future and ask @CarlosJCD  to create and follow up on this bug.

Good Job!

PR #114, opened 2023-11-17
KIN-300 File Validations

Mauriciocr207:
[KIN-300](https://loopcrack.atlassian.net/browse/KIN-300) Medium [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

It is needed to validate the files given by the users through the CtrlApiFile

### Solution

Define and integrate validations rules on the files given by the users

### Requirements checklist

- [x] There must be a File validations object
- [x] The file validation must be integrated on the files API
- [x] Must validate only one file at once according the way the CtrlApiFiles receives the files
- [x] There is a class with set of file validations rules defined and used into the validation services provided by codeigniter.

### Testing

Recommended testing by following:

- [x] Rememeber do `npm install` and `npm run dev:admin` and run the project on production mode
- [x] Go to the controller CtrlTestFiles, and you can see that there is some properties on the filepondConfiguration that are commented, that is in order you can test only the validation from server and filepond doesn´t interfere with the validation
- [x] put your screens like following:
image
- [x] Now you can use the following image to upload on filepond box and test this pull request:
- specifications:
   - size: 158 Kb
   - dims: 1920 x 1080
   - ext: .png
   - mime type: image/png

- [x] With the specification provided you know how to modify the validation rules on CtrlTestFiles, to prove every rule on validation.
- [x] For example, if you want to prove the maxSize rule, you can set `maxSize[100,kb]` in order the validation fails, because of the file size is 158 kb. you can set `maxSize[200,kb]` in order the validation pass.
- [x] you can do the above with the other validation rules.

## Demo

[KIN-300]: https://loopcrack.atlassian.net/browse/KIN-300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/Libraries/fileValidation/FileRules.php
+++ b/app/Libraries/fileValidation/FileRules.php
@@ -1,0 +1,162 @@
+<?php
+
+namespace App\Libraries\fileValidation;
+
+use CodeIgniter\Files\File;
+use Config\Mimes;
+
+class FileRules
+{
+    /**
+     * Verifies if the file's size in given unit is no larger than the parameter.
+     *
+     * @param File  $file   the instance of file
+     * @param mixed $params params passed through validation rules
+     */
+    public function maxSize(File $file, $params)
+    {
+        if ($file === null) {
+            return false;
+        }
+
+        $params = explode(',', $params);
+
+        $size = $file->getSizeByUnit(strtolower($params[1]));
+
+        $accepted = $params[0];
+
+        return $accepted > $size;
+    }
+
+    /**
+     * Verifies if the file's size in given unit is larger than the parameter.
+     *
+     * @param File  $file   the instance of file
+     * @param mixed $params params passed through validation rules
+     */
+    public function minSize(File $file, $params)
+    {
+        if ($file === null) {
+            return false;
+        }
+
+        $params = explode(',', $params);
+
+        $size = $file->getSizeByUnit(strtolower($params[1]));
+
+        $accepted = $params[0];
+
+        return $accepted < $size;
+    }
+
+    /**
+     * Uses the mime config file to determine if a file is considered an "image",
+     * which for our purposes basically means that it's a raster image or svg.
+     *
+     * @param File  $file   the instance of file
+     * @param mixed $params params passed through validation rules
+     */
+    public function isImage(File $file, $params)
+    {
+        $params = explode(',', $params);
+
+        if ($file === null) {
+            return false;
+        }
+
+        // We know that our mimes list always has the first mime
+        // start with `image` even when then are multiple accepted types.
+        $type = Mimes::guessTypeFromExtension($file->getExtension()) ?? '';
+
+        return ! (mb_strpos($type, 'image') !== 0);
+    }
+
+    /**
+     * Checks to see if an uploaded file's mime type matches one in the parameter.
+     *
+     * @param File  $file   the instance of file
+     * @param mixed $params params passed through validation rules
+     */
+    public function mimeIn(File $file, $params)
+    {
+        $params = explode(',', $params);
+
+        if ($file === null) {
+            return false;
+        }
+
+        return ! (! in_array($file->getMimeType(), $params, true));
+    }
+
+    /**
+     * Checks to see if an uploaded file's extension matches one in the parameter.
+     *
+     * @param File  $file   the instance of file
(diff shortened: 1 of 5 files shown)


app/Libraries/fileValidation/FileRules.php
```diff
@@ -0,0 +1,162 @@
...
+        $fileWidth  = $info[0];
+        $fileHeight = $info[1];
+
+        return ! ($fileWidth > $allowedWidth || $fileHeight > $allowedHeight);
+    }
+
+    /**
+     * Check if the array has the maximum items according to the max number passed on param
+     *
+     * @param File  $file   the instance of file
```
Oswaldo-Chan, line 138: This parameter doesn't correspond to what the function actually accepts; it should be an array called 'files' representing the number of files

app/Libraries/fileValidation/FileRules.php
```diff
@@ -0,0 +1,162 @@
...
+        $maxFiles = $params;
+        $numFiles = count($files);
+
+        return ! ($numFiles > $maxFiles);
+    }
+
+    /**
+     * Check if the array has the minimum items according to the min number passed on param
+     *
+     * @param File  $file   the instance of file
```
Oswaldo-Chan, line 152: This parameter doesn't correspond to what the function actually accepts; it should be an array called 'files' representing the number of files

Oswaldo-Chan (commented):
Hi @Mauriciocr207 I left you some comments in the `FileRules.php` file

CheloChidoone (changes requested):
@Mauriciocr207 The file extension validations are not working correctly, testing with Carlos we were able to upload different files that were not images like .csv and .zip, please check that out.

CC: @CarlosJCD

CarlosJCD (changes requested):
Hi @Mauriciocr207 while reviewing this PR we detected that although the files size are validated, when it comes to the files extensions some extension passes the validations even when they are not supposed to. We tested with a .csv and a .r as you can see in the image.
Screenshot 2023-11-18 at 14 43 59

Also, for some reason when I upload some kind of files [the page instantly refreshes when the upload reaches 100%](https://drive.google.com/file/d/1O5ERD3GCF0fwy4WwyY0psLchs-RWHl22/view?usp=sharing). This happened with .zip, .docx and some .pdf files. I am not sure if it is a problem with my environment but I resolved that it was worth to mention in the case you could knew why this happens.
 

CC: @CheloChidoone

Mauriciocr207:
> @Mauriciocr207 The file extension validations are not working correctly, testing with Carlos we were able to upload different files that were not images like .csv and .zip, please check that out.
> 
> CC: @CarlosJCD

I change the validation on CtrlApiFiles, please, check again. There was problems on validatino of files that are not uploaded by chunks.

Mauriciocr207:
> Hi @Mauriciocr207 while reviewing this PR we detected that although the files size are validated, when it comes to the files extensions some extension passes the validations even when they are not supposed to. We tested with a .csv and a .r as you can see in the image. Screenshot 2023-11-18 at 14 43 59
> 
> Also, for some reason when I upload some kind of files [the page instantly refreshes when the upload reaches 100%](https://drive.google.com/file/d/1O5ERD3GCF0fwy4WwyY0psLchs-RWHl22/view?usp=sharing). This happened with .zip, .docx and some .pdf files. I am not sure if it is a problem with my environment but I resolved that it was worth to mention in the case you could knew why this happens.
> 
> CC: @CheloChidoone

Sorry, I didn´t mention this but, that is due to the Browser sync. It reload the page sometimes. You could try using a virtual host.

app/Libraries/fileValidation/FileRules.php
Mauriciocr207: I did the corrections

CarlosJCD (approved):
Nice, now it works fine.

CarlosMay7 (approved):

CheloChidoone (approved):
Good Job @Mauriciocr207, It works well now.

Oswaldo-Chan (approved):

PR #120, opened 2023-11-20
KIN-99 add product image carousel

CarlosMay7:
[KIN-99](https://loopcrack.atlassian.net/browse/KIN-99) Large [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

The site's product page needs to show customers some images of the product in an attractive way.

### Solution

Create an image carousel to show different images about the product offering different ways to approach it.

### Requirements checklist

- [x] It must allow the user to swipe between images on the carousel
- [x] When it's on desktop display it must show an on-hover zoom over the gallery´s main image
- [x] When focusing on an image, this gets on a bigger resolution on top of the carousel
- [x] When it's on mobile display and clicked on the active slide it must show the same image in a bigger display and show a dark background

### Testing

- [x] Manually checked the responsiveness
- [x] Manually checked gestures like swiping and clicking

The LightGallery license key is in the resources channel in discord. You just have to add it in the licensekey value in light.js

## Demo

[Carousel Demo](https://drive.google.com/file/d/1FVvKuDjyGTwDtlR-6oxluf8DkrtRUUq1/view?usp=sharing)

[KIN-99]: https://loopcrack.atlassian.net/browse/KIN-99?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Cristianpan (commented):
Please specify where we should put the license in order to correctly test your implementation.

CC: @CarlosJCD

CarlosJCD (commented):
Is it supposed that the image should be changed when sliding with the images?, if so it doesnt happen.

Also we couldnt figure it out the purpose of the license key, we tested with and without the license key and we did not detected any changes on the carousel functionality? 

CC: @Cristianpan

LUGAMAFE (changes requested):
Can we add arrows to the glide carousel desktop version like this one @CarlosMay7 : 
image

Where are all this options too @CarlosMay7 in the mobile version for lightgalleryjs? 

image

@CarlosJCD and @Cristianpan i assume the .env part will be part of the BE integration, correct @CarlosMay7?

CheloChidoone (changes requested):
Good Job @CarlosMay7, just check the comments you already have from everyone and a couple of nitpicks from me is to remove your changes in .env.example since the license key is not going to be required there and add a little more space between the carousel and product title on mobile views

LUGAMAFE:
> Good Job @CarlosMay7, just check the comments you already have from everyone and a couple of nitpicks from me is to remove your changes in .env.example since the license key is not going to be required there and add a little more space between the carousel and product title on mobile views

@CheloChidoone the licence will be required in the future! if not, the plugin will not work in production

CheloChidoone:
> > Good Job @CarlosMay7, just check the comments you already have from everyone and a couple of nitpicks from me is to remove your changes in .env.example since the license key is not going to be required there and add a little more space between the carousel and product title on mobile views
> 
> @CheloChidoone the licence will be required in the future! if not, the plugin will not work in production

I know @LUGAMAFE I was just saying that May showed me the license only needs to be set in the .js file but it isn't necessary for it to be on the .env, that's what May told me so I just wrote it it my comment to remind him. do you think it does need to be in the .env?

Mauriciocr207 (commented):
Captura de pantalla (1477)

When you open the image, there is some errors in console. We make npm install and npm run dev:public, but the errors persist.
CC: @FernandoJoachin

FernandoJoachin (changes requested):
When I'm in tablet mode and try to swipe, the image automatically expands without me having clicked to do so.
CC: @Mauriciocr207

CarlosMay7:
> Can we add arrows to the glide carousel desktop version like this one @CarlosMay7 : image
> 
> Where are all this options too @CarlosMay7 in the mobile version for lightgalleryjs?
> 
> image
> 
> @CarlosJCD and @Cristianpan i assume the .env part will be part of the BE integration, correct @CarlosMay7?

I had a problem during the implementation of the controls in the mobile view because the lightgalley library has a different way of importing some things related to those toolbar and navigation icons I tried to solve but I couldn't so I decided to hide it and just show the thumbnails.
The problem is that when it tries to find a relative route to some fonts and an image in the node modules (place from where I am importing the CSS code) it can't find it.
Now about the .env part, we are still thinking about the best option.

CheloChidoone:
Thanks for the initial corrections @CarlosMay7, good job.

@LUGAMAFE I found some bugs with the lightgallery options you added:

- The light gallery opens up in desktop sizes not just mobile
- The close option button 'X' is not appearing in mobile, just in desktop
- The minimize option causes the page to get stuck and prevents scrolling 
image

CheloChidoone (approved):

PR #122, opened 2023-11-20
KIN-278 delete emails functionality email page

FernandoJoachin:
[KIN-278](https://loopcrack.atlassian.net/browse/KIN-278) Small [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

As an user of the admin panel, I wish to delete an email record so that in case of being needed it can be deleted permanently.

### Solution

Make sure that when you click the delete button, the email in the database is deleted

### Requirements checklist

- [x] It must delete the record from the database
- [x] It must return a response indicating success or failure on deleting the email record.

### Testing

- [x] Click the delete button
- [x] Verify that the record was deleted in the database

[KIN-278]: https://loopcrack.atlassian.net/browse/KIN-278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Mauriciocr207 (approved):

CarlosJCD (approved):

app/Controllers/CtrlAdminEmail.php
```diff
@@ -24,7 +24,16 @@ public function viewSpecificEmails($id)
 
     public function deleteEmail()
     {
-        $isDeleted = true;
+        $POST       = $this->request->getPost();
+        $emailId    = $POST['emailId'];
```
Cristianpan, line 28:
```suggestion
        $emailId       = $this->request->getPost('emailId');
```
You can get the email id directly from the post and it would be good if you use a try catch for database exception handling

Cristianpan (approved):

LUGAMAFE (approved):

CheloChidoone (changes requested):
Good job @FernandoJoachin but we found a bug, once the confirm button is clicked on the delete modal you should block the button because we can click on it many times and even though the delete works it returns an error message because we click it many times, please check that before merging.

CheloChidoone (approved):

CarlosMay7 (approved):

PR #143, opened 2023-11-29
KIN-219 Add delete category functionality without files management

Cristianpan:
[KIN-219](https://loopcrack.atlassian.net/browse/KIN-219) Small [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

The admin user needs to delete a category so that in case there is no longer needed it can discarded by the system

### Solution

Implement the delete category functionality

### Requirements checklist

- [x] It must delete the register from the database

For the moment, the files inputs will be ignored and when storing on the database, it will be a hardcoded value.
NOTES: The files functionality will be integrated late
### Testing

Manual testing
- [x] Select a category and delete it from the administration panel and check that the category data has been removed from the database along with its associated tags and files.

## Demo

[KIN-219]: https://loopcrack.atlassian.net/browse/KIN-219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/Controllers/CtrlCategory.php
```diff
@@ -82,21 +82,26 @@ public function updateCategory($id)
...
-        if ($isDeleted) {
-            $response = [
-                'title'   => 'Eliminación exitosa',
-                'message' => 'Se ha elimnado la categoría correctamente',
-                'type'    => 'success',
-            ];
-        } else {
+        $categoryId = $this->request->getPost('categoryId') ?? '';
+
+        (new CategoryModel())->deleteCategory($categoryId);
```
Mauriciocr207, line 87: Here is not neccesary verify if the register was deleted and send a response according to that. Is not neccesary to consider where the delete return `false`?

Oswaldo-Chan (approved):
👍 approved

Mauriciocr207 (approved):

CarlosMay7 (commented):
It works fine but when we checked the database the products related to the deleted category were still there when those were supposed to be deleted as well. Please check it out

Cristianpan:
> It works fine but when we checked the database the products related to the deleted category were still there when those were supposed to be deleted as well. Please check it out

That is because of how the test categories have been created, if you look at the seeders you can see that there are only 3 files created and those same files are associated to any entity that needs them, then when you delete the category, because of these associations between files, is what "fails" however in a normal flow that wouldn't happen, you can try creating a new category and you will see that everything works normally.

app/Models/CategoryModel.php
```diff
@@ -38,6 +39,23 @@ public function createCategory(array $categoryData)
         }
     }
 
+<<<<<<< HEAD
```
LUGAMAFE, line 42: @Cristianpan beware of these errors when resolving merge conflicts
```suggestion

```

app/Models/CategoryModel.php
```diff
@@ -38,6 +39,23 @@ public function createCategory(array $categoryData)
...
+
+            $fileModel    = new FileModel();
+            $imageDeleted = $fileModel->delete($categoryData['categoryImageId']);
+            $iconDeleted  = $fileModel->delete($categoryData['categoryIconId']);
+
+            if (! $iconDeleted || ! $imageDeleted || ! $categoryDeleted) {
+                throw new Exception('No se ha podido eliminar la categoría');
+            }
+
+=======
```
LUGAMAFE, line 58: same
```suggestion
```

app/Models/CategoryModel.php
```diff
@@ -60,6 +78,7 @@ public function updateCategory(string $categoryId, array $categoryData)
             }
 
             (new CategoryModel())->update($categoryId, $categoryData);
+>>>>>>> 51ecbe590e18479386d7c3b8d914ab2236ef7dad
```
LUGAMAFE, line 81: same
```suggestion
```

LUGAMAFE (changes requested):

app/Models/CategoryModel.php
LUGAMAFE, line 42: @Cristianpan beware of these errors when resolving merge conflicts, you didn't choose which version to use, right
```suggestion
```

LUGAMAFE (changes requested):

app/Models/CategoryModel.php
Cristianpan: Is right, I didn't see that conflict,  so sorry.

CheloChidoone (approved):
Good Job @Cristianpan

app/Controllers/CtrlCategory.php
```diff
@@ -132,4 +132,28 @@ public function updateCategory(string $categoryId)
...
+            $categoryId = $this->request->getPost('categoryId') ?? '';
+
+            (new CategoryModel())->deleteCategory($categoryId);
+            $response = [
+                'title'   => 'Eliminación exitosa',
+                'message' => 'La categoría ha sido eliminada correctamente',
+                'type'    => 'success',
+            ];
+
+            return redirect()->to('admin/categorias')->with('response', $response);
```
CarlosMay7, line 148: Maybe you can move his line outside of the try catch and everything should work fine

app/Controllers/CtrlCategory.php
```diff
@@ -132,4 +132,28 @@ public function updateCategory(string $categoryId)
...
+        } catch (Throwable $th) {
+            $response = [
+                'title'   => 'Oops! Ha ocurrido un error.',
+                'message' => 'Ha ocurrido un error al eliminar los datos de la categoría, por favor intente nuevamente.',
+                'type'    => 'error',
+            ];
+
+            return redirect()->to('admin/categorias')->with('response', $response);
```
CarlosMay7, line 156: Because is the same statement as before

app/Models/CategoryModel.php
```diff
@@ -67,4 +68,27 @@ public function updateCategory(string $categoryId, array $categoryData)
...
+
+            $fileModel    = new FileModel();
+            $imageDeleted = $fileModel->delete($categoryData['categoryImageId']);
+            $iconDeleted  = $fileModel->delete($categoryData['categoryIconId']);
+
+            if (! $iconDeleted || ! $imageDeleted || ! $categoryDeleted) {
+                throw new Exception('No se ha podido eliminar la categoría');
+            }
+
+            $this->db->transComplete();
```
CarlosMay7, line 87: Check the suggestion you did about the change from transComplete to transCommit

CarlosMay7 (changes requested):
Hey @Cristianpan, we checked and everything looks good overall, just a few comments. 

Also we have a question, shouldn't we delete the category tags along with the category?

CC: @FernandoJoachin

Cristianpan:
> Hey @Cristianpan, we checked and everything looks good overall, just a few comments.
> 
> Also we have a question, shouldn't we delete the category tags along with the category?
> 
> CC: @FernandoJoachin

I've made the suggested changes, but about deleting category tags, it works correctly for me and I've not made it delete through the code because I'm delegating that action to the Db. I'd like that you to confirm if you are using a proper configuration for the Db (the eginee innodb). Likewise if anyone can confirm whether or not it is working correctly for them, that would be nice.

FernandoJoachin (approved):

CarlosMay7 (approved):
Nice work, thank you

CarlosJCD (approved):
Nice work!, remember to resolve comments before doing merge

LUGAMAFE (approved):

PR #144, opened 2023-12-01
KIN-210: Edit measurement solution functionality

CarlosJCD:
[KIN-210](https://loopcrack.atlassian.net/browse/KIN-210) Small [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

Update / change the already registered data of a measurement solution.

### Solution

Implement the edit logic for registered measurement solutions

### Requirements checklist

- [x] It must validate the inputs to not be empty

- [x] In case the inputs are correct, it must update the register of the measurement solution on the database

- [x] In case there are errors, it must return them to the view

- [x] It must return a response to the view indicating success or error on persisting the changes.

NOTES: The files functionality will be integrated on a later Issue

### Testing

- [x] On the edit measurement solution view, modify the name or  description on the form. Then click on the submit button and observe how it redirects to the measurement solutions page with a response indicating success and the new data updated on the table.

## Demo

[KIN-210]: https://loopcrack.atlassian.net/browse/KIN-210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/Views/admin/typeSolutions/solutionsFormTemplate.php
```diff
@@ -70,4 +70,6 @@
     </div>
     <!-- end col -->
 </div>
+
+<input type="hidden" name="msId" value="<?php old('msId') ?? $measurementSolution['msId'] ?? ''; ?>" >
```
CheloChidoone, line 74: This input shouldn't be necessary if all you want from it is the `msId` value because you should already be getting that from the route for the update

app/Controllers/CtrlSolution.php
```diff
@@ -67,19 +67,33 @@ public function createSolution()
...
     }
 
-    public function updateSolution($id)
+    public function updateSolution(string $msId)
     {
-        $isUpdated = true;
-        if ($isUpdated) {
+        try {
+            $msData          = $this->request->getPost();
+            $msData['msId']  = $msId;
```
CheloChidoone, line 74: Why are you overwriting the $msId that you receive as a parameter on this function, this shouldn't be necessary

CheloChidoone (changes requested):
@CarlosJCD Good Job, it works well but we found some things that you added without it being necessary, check how update functions for other elements have been done for reference and check these comments.

CC: @Cristianpan

Mauriciocr207 (commented):
I left you some comments
CC: @Oswaldo-Chan

app/Controllers/CtrlSolution.php
LUGAMAFE, line 74:
```suggestion
```

there is no need for this
follow the pattern recommended by mau

`$msModel->updateData($msId, $msData);`

LUGAMAFE (approved):
Good Job @CarlosJCD

Mauriciocr207 (approved):

Cristianpan (approved):

Oswaldo-Chan (approved):

CheloChidoone (approved):
Good Job @CarlosJCD!

PR #145, opened 2023-12-11
KIN-289 separate endpoints for confirm account and change password features

FernandoJoachin:
[KIN-289](https://loopcrack.atlassian.net/browse/KIN-289) Medium [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

It is necessary to separate the confirm account functionality and reset account functionality in two new endpoints.

### Solution

Separate the confirm account functionality and reset password functionality in two different endpoints, so that each endpoint has its own routes and views 

### Requirements checklist

- [x] The confirm account view, implemented logic and it must be on its own endpoint.
- [x] The change password view, implemented logic and it must be on its own endpoint.

### Testing

- [x] Please go to http://kinub.cm/password_reset and try adding an account that hasn't been confirmed yet.
- [x] Go to http://kinub.cm/password_reset and try adding a confirmed account. Then, go to Mailtrap and click on the 'reset password' button. Next, attempt to enter a token that doesn't exist in the URL or a token from someone who hasn't confirmed their account. This action should redirect you to http://kinub.cm/password_response with an error response.
- [x] Create a new user. Then, go to Mailtrap and click on the 'set password' button. Afterward, try entering a token that doesn't exist in the URL or a token from someone who has already confirmed their account. This action should redirect you to http://kinub.cm/password_response with an error response.

[KIN-289]: https://loopcrack.atlassian.net/browse/KIN-289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/Controllers/CtrlLogin.php
```diff
@@ -60,51 +79,68 @@ public function passwordReset($token)
...
 
-            $userTokenModel = new UserTokenModel();
-            $userToken      = $userTokenModel->where('userToken', $token)->first();
+            $user = TokenUtils::getUserWithToken($token);
+
+            $changePasswordValidation->existUserWithToken($user);
 
             $userModel = new UserModel();
-            $user      = $userModel->find($userToken['userId']);
+            $userModel->updatePassword($user['userId'], $data['password'], 1);
```
LUGAMAFE, line 90: And now you need to remove the used token! you are not deleting used tokens

app/Controllers/CtrlLogin.php
```diff
@@ -60,51 +79,68 @@ public function passwordReset($token)
...
+                throw new InvalidInputException($changePasswordValidation->getErrors());
             }
+
+            $user = TokenUtils::getUserWithToken($token);
+
+            $changePasswordValidation->existUserWithToken($user);
+
+            $userModel = new UserModel();
+            $userModel->updatePassword($user['userId'], $data['password'], 0);
+
```
LUGAMAFE, line 127: Same here you need to  remove the used token!

LUGAMAFE (changes requested):
@FernandoJoachin please talk with @CarlosJCD you have several code logic problems

app/Utils/TokenUtils.php
```diff
@@ -19,4 +19,18 @@ public static function generateToken($userId)
 
         return $tokenData['userToken'];
     }
+
+    public static function getUserWithToken($token)
+    {
+        $userTokenModel = new UserTokenModel();
+
+        return $userTokenModel->join('users', 'users.userId = user_tokens.userId')->where('userToken', $token)->first();
```
Cristianpan, line 27: I think that those methods should be in the UserTokenModel and not such as an util

app/Validation/ChangePasswordValidation.php
```diff
@@ -25,11 +25,23 @@ class ChangePasswordValidation extends BaseValidation
...
+
+    public function accessValidation($isConfirmed, $isAuthorized)
+    {
+        if ($isConfirmed !== $isAuthorized) {
+            $message = '';
+            if ($isAuthorized) {
+                $message = 'Parece que no cuenta con el permiso para restablecer su contraseña';
+            } else {
+                $message = 'Parece que no cuenta con el permiso para establecer su contraseña';
+            }
```
Mauriciocr207, line 42: I think you can separate this function on two, to be more descriptive for example:

```php
public function hasAccessToSetPassword($isConfirmed)
    {
        if ($isConfirmed !== USER_IS_NOT_CONFIRMED) {
            throw new Exception('Parece que no cuenta con el permiso para establecer su contraseña');
        }

        return true;
    }

public function hasAccessToResetPassword()
    {
        if ($isConfirmed !== USER_IS_CONFIRMED) {
            throw new Exception('Parece que no cuenta con el permiso para restablecer su contraseña');
        }

        return true;
    }
```

app/Controllers/CtrlLogin.php
```diff
@@ -21,34 +22,52 @@ public function viewPasswordEmail(): string
...
+        return view('login/PasswordResponse');
+    }
+
+    public function viewPasswordReset($token)
     {
         try {
             $changePasswordValidation = new ChangePasswordValidation();
+            $user                     = TokenUtils::getUserWithToken($token);
+            $changePasswordValidation->existUserWithToken($user);
+            $changePasswordValidation->accessValidation($user['confirmed'], '1');
```
Mauriciocr207, line 36: And when you separate this functionality you can use the constants `USER_IS_NOT_CONFIRMED` like:
```php
$changePasswordValidation->hasAccessToSetPassword($user['confirmed'], USER_IS_NOT_CONFIRMED);
```

app/Controllers/CtrlLogin.php
```diff
@@ -21,34 +22,52 @@ public function viewPasswordEmail(): string
...
 
+    public function viewPasswordSet($token)
+    {
+        try {
+            $changePasswordValidation = new ChangePasswordValidation();
+            $user                     = TokenUtils::getUserWithToken($token);
             $changePasswordValidation->existUserWithToken($user);
-
-            $isAccountConfirmed = $user['confirmed'];
+            $changePasswordValidation->accessValidation($user['confirmed'], '0');
```
Mauriciocr207, line 58: And here:
```php
$changePasswordValidation->hasAccessToResetPassword($user['confirmed'], USER_IS_CONFIRMED);
```

Mauriciocr207 (changes requested):

Cristianpan (changes requested):
Please considere the changes.

CC: @Mauriciocr207

Oswaldo-Chan (approved):

Oswaldo-Chan (approved):

CheloChidoone (approved):
Good Job @FernandoJoachin, thanks for adding the validations and changes requested, on our tests expired token validation and the tokens being deleted after being used work correctly, well done.

CC: @Oswaldo-Chan

app/Validation/ChangePasswordValidation.php
```diff
@@ -21,15 +26,55 @@ class ChangePasswordValidation extends BaseValidation
...
+            throw new Exception('Parece que no cuenta con el permiso para establecer su contraseña');
+        }
+
+        return true;
+    }
+
+    public function hasAccessToResetPassword()
+    {
+        if ($this->userWithToken['confirmed'] !== self::USER_IS_CONFIRMED) {
+            throw new Exception('Parece que no cuenta con el permiso para restablecer su contraseña');
```
Mauriciocr207, line 77: I have a doubt, in the `hasAccessToSetPassword()`, you need that user is confirmed right? then you can make a positive sentence on if statement, and in the `hasAccessToResetPassword()` function occurs similar, then:
```suggestion
    public function hasAccessToSetPassword()
    {
        if ($this->userWithToken['confirmed'] === self::USER_IS_CONFIRMED) {
            throw new Exception('Parece que no cuenta con el permiso para establecer su contraseña');
        }

        return true;
    }

    public function hasAccessToResetPassword()
    {
        if ($this->userWithToken['confirmed'] === self::USER_IS_NOT_CONFIRMED) {
            throw new Exception('Parece que no cuenta con el permiso para restablecer su contraseña');
```

app/Controllers/CtrlEmail.php
```diff
@@ -136,7 +137,7 @@ public function sendEmailToResetPassword()
 
         try {
             if (! $validateEmail->validateInputs($data)) {
-                throw new Exception();
+                throw new InvalidInputException($validateEmail->getErrors());
```
Mauriciocr207, line 140: Here remember put the `$validateEmail` and the `$data` into the try-catch, you can not see it like a change, but if you check your code, those variables are out the try-catch

Mauriciocr207 (approved):
Here there is some suggestions

app/Models/UserModel.php
```diff
@@ -30,4 +32,33 @@ protected function hashPassword(array $data)
...
+                ]);
+            }
+
+            $userTokenModel = new UserTokenModel();
+
+            $responseDelete = $userTokenModel->where('userId', $userId)->delete();
+            if (! $responseDelete) {
+                throw new Exception();
+            }
+            $this->db->transComplete();
```
CarlosMay7, line 57: Here check the comments @Cristianpan made about using transCommit instead of transComplete

CarlosMay7 (approved):
Nice work, just check the comment I left you 

And also check the little conflict about the token generator with @Cristianpan

app/Controllers/CtrlLogin.php
```diff
@@ -11,6 +12,9 @@
 
 class CtrlLogin extends BaseController
 {
+    public const USER_IS_NOT_CONFIRMED = '0';
+    public const USER_IS_CONFIRMED     = '1';
+
```
Cristianpan, line 17: This is fine, but we have a file with all the constants we need for this project so I think it would be a good idea to add them to that file, especially since I see it used here and also in a validator. The file path is `/app/Config/Constants.php`.

app/Models/UserModel.php
```diff
@@ -30,4 +32,33 @@ protected function hashPassword(array $data)
 
         return $data;
     }
+
+    public function updatePassword($userId, $newPassword, $isConfirmed)
+    {
+        try {
+            $this->db->transStart();
```
Cristianpan, line 39:
```suggestion
            $this->db->transException(true)->transStart();
```
Make this change please for the enable that the db throws the exceptions

app/Config/Constants.php
```diff
@@ -100,3 +100,12 @@
...
 define('FILES_UPLOAD_DIRECTORY', './uploads/');
+
+/*
+ | --------------------------------------------------------------------------
+ | User account status
+ | --------------------------------------------------------------------------
+ /
+ */
+define('USER_IS_NOT_CONFIRMED', '0');
+define('USER_IS_CONFIRMED', '1');
```
CarlosJCD, line 111: I am not really sure if it is ok to define as a global constant a value that is currently only used in the ctrlLogin class. I believe that it would be better to manage this kind of constants as private attributes of the class so that it can have the right scope. Maybe its just my impression, what do you think about this? @Cristianpan , @Mauriciocr207 , @CheloChidoone , @LUGAMAFE

CarlosJCD (changes requested):
Nice work!, just check out the comments i left you.

app/Config/Constants.php
LUGAMAFE: yeah @CarlosJCD that is correct if is not being used i other parts should be a class constant

Cristianpan: Hmm ok but I had told Joachin to put it there since I saw that he was using it in a controller and a validator, also thinking that those constants could be reused. So I'd think that putting them in that file would be fine but I don't know what others think.

FernandoJoachin: So, how do I do it?"

> I am not really sure if it is ok to define as a global constant a value that is currently only used in the ctrlLogin class. I believe that it would be better to manage this kind of constants as private attributes of the class so that it can have the right scope. Maybe its just my impression, what do you think about this? @Cristianpan , @Mauriciocr207 , @CheloChidoone , @LUGAMAFE

So, how do I do it?"

app/Views/login/PasswordResponse.php
```diff
@@ -0,0 +1,65 @@
...
+
+            </div>
+            <!-- end container -->
+        </div>
+        <!-- end auth page content -->
+
+    </div>
+    <!-- end auth-page-wrapper -->
+
+    <?= $this->include('partials/vendor-scripts') ?>
```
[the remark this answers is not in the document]
CarlosJCD: @FernandoJoachin please check this out so that I can accept your PR

app/Config/Constants.php
CarlosJCD: If you are using it in more than one class then leave it like that!, my observation was because I thinked that these constants where used only on the controller.

LUGAMAFE (approved):

CarlosJCD (approved):

PR #147, opened 2023-12-11
KIN-290 Add resend email to confirm account functionality

Cristianpan:
[KIN-290](https://loopcrack.atlassian.net/browse/KIN-290) Medium [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem
It is needed to generate a feature on the admin panel that allows the admin to send again the confirm account email for unconfirmed accounts.

### Solution

Implement functionality that allows the admin user to resend an email to confirm the account.

### Requirements checklist

- [x] There must be a button for the users page and edit users views that must allow to resend the confirmation email to the unconfirmed account.
- [x] It must be displayed a modal alert indicating success on resending the new confirm account or if there was an error in the process.
- [x] When is triggered this button, the following must happen:
- [x] The last created token for the account must be deleted.
- [x] A new token wit a new expiration date (current date + 24 hours) must be generated and registered on the database token table.
- [x] The confirm account must be sent again with this new token.

### Testing

Modify the "confirmed" attribute of a registered user and do the following and login to the admin panel with the superadmin account.
- [x] Try to resend the account confirmation email from the administration panel and verify in mailtrap that you have received a new account confirmation email. Also check that the token has been updated in the database.
- [x] Try to resend the account confirmation email from the user edit view and check the same as in the previous point.

## Demo
image
image

[KIN-290]: https://loopcrack.atlassian.net/browse/KIN-290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CheloChidoone (changes requested):
Good job Cristian, it works well but when testing it from the button you added on the edit user page the confirmation modal does not show up, it just looks like it just recharged the page and nothing else appears even though it did work, just check that bug

CC: @Oswaldo-Chan

app/Controllers/CtrlUser.php
```diff
@@ -140,6 +141,35 @@ public function updateUser($id)
...
+
+            return redirect()->back()->with('response', $response);
+        } catch (Throwable $th) {
+            $response = [
+                'title'   => 'Oops! Ha ocurrido un error',
+                'message' => 'Algo ha salido mal mientras se enviaba el email, por favor intente nuevamente',
+                'type'    => 'error',
+            ];
+
+            return redirect()->back()->with('response', $response);
```
FernandoJoachin, line 169: On lines 161 and 169, you have the same code. Take it out of the try-catch block to avoid code duplication. It should work
CC: @CarlosJCD

app/Models/UserTokenModel.php
```diff
@@ -13,4 +16,34 @@ class UserTokenModel extends Model
...
+                'userToken'       => substr((new Uuid())->uuid3(), 0, 13),
+                'tokenExpiryDate' => date('y-m-d', strtotime(' +1 day')),
+                'userId'          => $userId,
+            ];
+
+            if ($this->insert($tokenData) === false) {
+                throw new Exception('Ha ocurrido un error al generar el nuevo token');
+            }
+
+            $this->db->transComplete();
```
FernandoJoachin, line 40: The same comment. Use 'transCommit' instead of 'TransComplete'

FernandoJoachin (commented):
Why did you remove TokenGenerator?

Cristianpan:
> Why did you remove TokenGenerator?

Because the utility could be replaced by the model events provided by CI in the models since that action is always performed before generating the token, then I considered that it would be better to delegate that action to the model event instead of creating a utility. 
You can read more about this in the CI models [here](https://www.codeigniter.com/user_guide/models/model.html#model-events)

Cristianpan:
> Why did you remove TokenGenerator?

Because replace the utility by the method in the model, where the token to be inserted is created directly there. I considered that it was better to have it as a utility.

Oswaldo-Chan (approved):

CarlosJCD (approved):
Excellent work!

PR #152, opened 2023-12-15
KIN-93 API Searchbar - Product Filters API - BackEnd

Mauriciocr207:
[KIN-93](https://loopcrack.atlassian.net/browse/KIN-93) Medium [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

It is needed to handle the frontend requests so that the filtered products can be retrieved from the server. 

### Solution

Implement a product filter API

### Requirements checklist

- [x] Provide an api filter to search by name between filtered products
- [x] The filters must validate query params received from the client request
- [x] It must implement the endpoints for the product filters functionality
- [x] It must retrieve the products with the requested filters

### Testing

- [x] In order to test this pr remember do `php spark migrate:refresh` and `php spark db:seed MainTestSeeder` to add more registers to test correctly this pr.
- [x] Second, to test the api you need to download and install [Insomnia](https://insomnia.rest/download). To get the project to test, you can check you loopcrack mail and acept the invitation. When you get into insomnia, go to the public api and make pull to get all changes already made on the api.

## Resources

- [Insomnia](https://insomnia.rest/)
- [documentation](https://loopcrack.atlassian.net/wiki/spaces/Kinub/pages/69107714/Product+Search+Filter+Api)

## Demo

[KIN-93]: https://loopcrack.atlassian.net/browse/KIN-93?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/Controllers/CtrlApiPublic.php
```diff
@@ -0,0 +1,132 @@
...
+            return $this->response->setStatusCode(Response::HTTP_INTERNAL_SERVER_ERROR, json_encode('Ha ocurrido un error en el filtro'));
+        }
+    }
+
+    /**
+     * just for get product tags from products asociated to a category
+     */
+    public function getProductTags()
+    {
+        $categoryIds      = $this->request->getGet('category');
```
CheloChidoone, line 114: Same, We should be consistent in the naming of the query params to make it easier, make it either `category` or `categories` for all endpoints

app/Controllers/CtrlApiPublic.php
```diff
@@ -0,0 +1,132 @@
...
+
+        return json_encode($categoryTags);
+    }
+
+    /**
+     * just get background image of a category
+     */
+    public function getCategoryImage()
+    {
+        $categoryId    = (string) $this->request->getGet('category');
```
CheloChidoone, line 126: Same, We should be consistent in the naming of the query params to make it easier, make it either `category` or `categories` for all endpoints

app/Controllers/CtrlApiPublic.php
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * TEST FUNCTION
+     *
+     * the following functions are going to be removed. They are just for testing
+     */
+    public function getFilteredProduct()
+    {
+        try {
+            $request      = $this->request;
+            $categories   = $request->getGet('categories');
```
CheloChidoone, line 71: We should be consistent in the naming of the query params to make it easier, make it either `category` or `categories` for all endpoints

CheloChidoone (changes requested):
Great Job @Mauriciocr207, it works awesome but please check these comments I left you and also update the api documentation because the names of the query params are different for a lot of them to what you actually did in the code so that gave us a lot of trouble testing until we found out the names of the params on the documentation are wrong.

CC: @Oswaldo-Chan

app/Controllers/CtrlApiPublic.php
Mauriciocr207: Here I name this with the query param `"category"` because we need only one category for get the image, but in the filter and search methods we need one o more categories

CarlosMay7 (commented):
Looks good overall but I have an issue related to the sorting because it doesn't work in either way, in addition to another problem, what is the difference between order and direction?

Mauriciocr207:
> Looks good overall but I have an issue related to the sorting because it doesn't work in either way, in addition to another problem, what is the difference between order and direction?

The order refers to the type of ordering by name or id. And direction can be ascendet or descendent.
Can you give more information about the problem?

app/Models/ProductModel.php
```diff
@@ -2,17 +2,107 @@
...
+            $whereQueries[] = "(products.productName LIKE '%{$search}%' OR products.productModel LIKE '%{$search}%')";
+        }
+
+        $builder->groupBy('productId');
+
+        if (empty($whereQueries)) {
+            return $this;
+        }
+        $whereQuery = implode(' AND ', $whereQueries);
+        $builder->where($whereQuery);
```
Cristianpan, line 74: Instead of creating `where in` statements manually, you can make use of the whereIn method provided by the CI query builder, and the same case with the like statement.

```suggestion
             // added categories filter
        if (! empty($categories)) {
            $builder->whereIn('categories.categorySlug', $categories);
        }
        // added category_tags filter
        if (! empty($categoryTags)) {
            $builder->whereIn('category_tags.categoryTagSlug', $categoryTags);
        }
        // added products_tags filter
        if (! empty($productTags)) {
            $builder->whereIn('product_tags.ptSlug', $productTags);
        }
        // added search filter
        if (! empty($search)) {
            $builder->like('products.productName', $search)->orLike('products.productModel', $search);
        }

        $builder->groupBy('productId');

        return $this;
```

app/Models/ProductModel.php
```diff
@@ -2,17 +2,107 @@
...
+     * @param string $direction asc or desc
+     *
+     * @return $this
+     */
+    public function order(string $order = 'asc')
+    {
+        $this->orderBy('productName', $order);
+
+        return $this;
+    }
```
Cristianpan, line 92: The name of this function could be confusing, I think a better name would be `orderByProductName`. And for the parameter, maybe a better descriptive name would be `orderType`.

Cristianpan (changes requested):

Cristianpan:
Likewise, check that the documentation corresponds to the current implementation, as well as what you've provided to test the api.

app/Models/ProductModel.php
```diff
@@ -2,17 +2,107 @@
...
+     * Create the query for filter products by categories, categoryTags, productTags and a search by productName or productModel
+     *
+     * @param array  $categories   all categories to filter
+     * @param array  $categoryTags all categoryTags to filter
+     * @param array  $productTags  all productTags to filter
+     * @param string $search       the product name or product model to filter
+     *
+     * @return $this
+     */
+    public function filterProducts(array $categories, array $categoryTags, array $productTags, string $search = '')
```
LUGAMAFE, line 34:
```php
class ProductModel extends Model
{
    // ... [resto del código] ...

    public function filterProducts(array $categories, array $categoryTags, array $productTags, string $search = '')
    {
        $builder = $this->buildBaseQuery();

        $this->applyFilter($builder, 'categories.categorySlug', $categories);
        $this->applyFilter($builder, 'category_tags.categoryTagSlug', $categoryTags);
        $this->applyFilter($builder, 'product_tags.ptSlug', $productTags);
        $this->filterBySearch($builder, $search);

        $builder->groupBy('productId');

        return $this;
    }

    // Resto de métodos...

    private function applyFilter($builder, string $fieldName, array $values)
    {
        if (!empty($values)) {
            $builder->whereIn($fieldName, $values);
        }
    }

    private function filterBySearch($builder, string $search)
    {
        if (!empty($search)) {
            $builder->groupStart()
                    ->like('products.productName', $search)
                    ->orLike('products.productModel', $search)
                    ->groupEnd();
        }
    }

    // ... [resto del código] ...
}```

LUGAMAFE (changes requested):
Nice @Mauriciocr207 couple nitpicks.
And please can you update de documentation and insomnia API 🙏

Cristianpan:
> All good @Mauriciocr207 just please check this small bug, and please update the insomnia contract with only the required query params. CC @Oswaldo-Chan image
> 
>  2024-01-09.10-51-45.mp4

About the bug, I think it doesn't return the name of some product because the new category tag you created hasn't been associated to some product. I've tried the same thing but associating it and it does return some results. 

image

And about Insomnia, it's "complicated" to update so I left you a video demo about how to update the API documentation in Insomnia.


https://github.com/loopcrack-org/Kinub/assets/91033040/43b1b060-973e-4789-a209-2f97aa6912dd

LUGAMAFE (approved):
Uff thanks for the info @Cristianpan yeap that worked for me! good job @Mauriciocr207 :)

Mauriciocr207:
> All good @Mauriciocr207 just please check this small bug, and please update the insomnia contract with only the required query params. CC @Oswaldo-Chan image
> 
>  2024-01-09.10-51-45.mp4

I updated the pr description now, but yes. As Cristian mentioned, it is necessary to make pull on insomnia api
About the bug, I have the same question, is there any product with that category?

Oswaldo-Chan (approved):

CheloChidoone (approved):

Cristianpan (approved):

Cristianpan (approved):

CarlosMay7 (approved):

PR #156, opened 2023-12-15
KIN-335: Files Management Functionality on Measurement Solutions CRUD

CarlosJCD:
[KIN-335](https://loopcrack.atlassian.net/browse/KIN-335) Medium [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

The CRUD operations for the measurement solutions entities does not handle files values.

### Solution

It is implemented the files management logic on the CRUD operations of the measurement solutions entities

### Requirements checklist

#### On Create:
- [x] It must validate the file inputs
   -  The icon input must receive only 1 file, with a max size of 1MB and the file can only have a .svg extension
   -  The icon input must receive only 1 file, with a max size of 2MB and the file can only have a .jpg, .jpeg or .png extension.
- [x] It must store the measurement solutions files on the server and files routes on the database. Finally link them in measurement solution record.
#### On Update:
- [x] If the measurement solution image and/or icon are changed, the new files must be stored on the server and their routes stored in the database, as well as link this new file records to the measurement solution record. Finally the old files must be deleted from the server and the database.

#### On Delete:
- [x] It must delete the measurement solutions files on the server and from the database.

### Testing

- [x] Create a new measurement solution record using the create view, you will see the new filepond inputs as well as its validation when you upload a file on the input. Also when the measurement solution is created you will see the stored files on the `/uploads` directory.
- [x] Edit an existing measurement solution file in the edit view. Observe how when a new file is uploaded, the old one is deleted and the new one is registered on the database and stored on the uploads directory
- [x] Delete an existing measurement solution and observe how the uploaded files are deleted.

## Demo

[video demo](https://drive.google.com/file/d/1P4vNMfw6zYSUM0Etrb3IsUcvcOSb3-I2/view?usp=drive_link)

[KIN-335]: https://loopcrack.atlassian.net/browse/KIN-335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/Controllers/CtrlSolution.php
```diff
@@ -98,36 +134,67 @@ public function updateSolution(string $msId)
...
+
             $response = [
                 'title'   => 'Eliminación exitosa',
-                'message' => 'Se ha elimnado la solución de medición correctamente',
+                'message' => 'Se ha eliminado la solución de medición correctamente',
                 'type'    => 'success',
             ];
 
-            return redirect()->to('/admin/soluciones')->with('response', $response);
+            return redirect()->to($this->MEASUREMENT_SOLUTIONS_BASE_ROUTE)->with('response', $response);
```
Cristianpan, line 161: Update this constant 
```suggestion
            return redirect()->to(self::MEASUREMENT_SOLUTIONS_BASE_ROUTE)->with('response', $response);
```

app/Controllers/CtrlSolution.php
```diff
@@ -98,36 +134,67 @@ public function updateSolution(string $msId)
...
         } catch (Throwable $th) {
             $response = [
                 'title'   => 'Oops! Ha ocurrido un error.',
                 'message' => 'Ha ocurrido un error al tratar de eliminar la solución de medición, por favor intente nuevamente.',
                 'type'    => 'error',
             ];
 
-            return redirect()->to('/admin/soluciones')->with('response', $response);
+            return redirect()->to($this->MEASUREMENT_SOLUTIONS_BASE_ROUTE)->with('response', $response);
```
Cristianpan, line 169: Update this constant 
```suggestion
            return redirect()->to(self::MEASUREMENT_SOLUTIONS_BASE_ROUTE)->with('response', $response);
```

app/Models/MeasurementSolutionModel.php
```diff
@@ -12,21 +12,87 @@ class MeasurementSolutionModel extends Model
...
+
+            $fileModel = new FileModel();
+
+            $fileModel->delete($msData['msIconId']);
+            $fileModel->delete($msData['msImageId']);
+
+            return $this->db->transCommit();
+        } catch (Throwable $th) {
+            throw $th;
+        }
```
Cristianpan, line 96: Please add the rollback sentence 
```suggestion
        } catch (Throwable $th) {
           $this->db->transRollback();
            throw $th;
        }
```

app/Models/MeasurementSolutionModel.php
```diff
@@ -12,21 +12,87 @@ class MeasurementSolutionModel extends Model
...
+            $fileModel = new FileModel();
+
+            $msData['msIcon']  = $fileModel->select('uuid')->where('fileId', $msData['msIconId'] ?? '')->first()['uuid'] ?? '';
+            $msData['msImage'] = $fileModel->select('uuid')->where('fileId', $msData['msImageId'] ?? '')->first()['uuid'] ?? '';
+
+            return $msData;
+        } catch (Throwable $th) {
+            throw $th;
+        }
+    }
```
Cristianpan, line 48: I consider that this try catch is unnecessary since you do nothing more than catch the same error and throw it, which would be the same as if there were no try catch.

Cristianpan (changes requested):
When server error messages are displayed, the form layout is deformed. 
image

And in the solution edit view the validation errors are not displayed and the filepond status is not preserved either.
[Video Demo](https://youtu.be/g-ggGFEubk0)

app/Controllers/CtrlSolution.php
```diff
@@ -2,14 +2,33 @@
...
+    private const MEASUREMENT_SOLUTIONS_EDIT_ROUTE   = self::MEASUREMENT_SOLUTIONS_BASE_ROUTE . 'editar/';
+
+    protected FileValidationConfig $fileConfig;
+
+    public function __construct()
+    {
+        $fileConfigBuilder = new FileValidationConfigBuilder(self::MEASUREMENT_SOLUTIONS_BASE_ROUTE);
+
+        $fileConfigBuilder->builder('msIcon')->isSVG()->minFiles(1)->maxFiles(1)->maxSize(1, 'MB')->build();
+        $fileConfigBuilder->builder('msImage')->isImage()->minFiles(1)->maxFiles(1)->maxSize(2, 'MB')->build();
```
FernandoJoachin, line 27: Add the maxDimensions function to both.

app/Models/MeasurementSolutionModel.php
```diff
@@ -12,17 +12,81 @@ class MeasurementSolutionModel extends Model
...
+
+    public function deleteMeasuremenSolution(array $msData)
+    {
+        try {
+            $this->db->transException(true)->transStart();
+
+            $fileModel = new FileModel();
+
+            $fileModel->delete($msData['msIconId']);
+            $fileModel->delete($msData['msImageId']);
```
FernandoJoachin, line 87: @Cristianpan Pan, I have the following question. Since TransException has been added, is it no longer necessary to validate if the delete query returned false to throw an exception?"

app/Controllers/CtrlSolution.php
Mauriciocr207: @FernandoJoachin in the case of svg, is not neccesary to set dimensions

app/Models/MeasurementSolutionModel.php
Cristianpan: It's correct because before if there was an error, the delete returned a false but with the use of `$this->db->transException(true)->transStart()` now it throws an exception if there is an error.

CheloChidoone (approved):
Good Job @CarlosJCD, just finish checking the comments that Cristian and Joachin left you

Oswaldo-Chan (approved):

app/Models/MeasurementSolutionModel.php
Mauriciocr207, line 87: I think this is incorrect. I think there is an incorrect relation on db, because you are deleting the entity measurement solution through the files. And I think that is an incorrect relation. Because of the way the db is already built, I think you need to delete the entity measurement solution first, and then delete the files. And I think you can change the db on fk actions, setting on delete set null and on update set null

Mauriciocr207 (changes requested):

app/Controllers/CtrlSolution.php
```diff
@@ -20,114 +39,170 @@ public function viewSolutions()
...
+
         try {
-            $newMeasurementSolutionData       = $this->request->getPost();
-            $measurementSolutionDataValidator = new SolutionValidation();
+            $msDataValidator = new SolutionValidation();
 
-            if (! $measurementSolutionDataValidator->validateInputs($newMeasurementSolutionData)) {
-                throw new InvalidInputException($measurementSolutionDataValidator->getErrors());
+            if (! $msDataValidator->validateInputs($msData)) {
+                throw new InvalidInputException($msDataValidator->getErrors());
```
Mauriciocr207, line 73: Here you are not validating the files. This could be like following:
```suggestion
            $fileValidationRules = $this->fileConfig->getCollectionFileValidationRules();
            $msDataValidator->addRules($fileValidationRules['rules'], $fileValidationRules['messages']);
            if (! $msDataValidator->validateInputs($msData)) {
                throw new InvalidInputException($msDataValidator->getErrors());
            }
```

app/Controllers/CtrlSolution.php
```diff
@@ -20,114 +39,170 @@ public function viewSolutions()
...
-            $msData = $this->request->getPost();
+        $msUpdatedData = $this->request->getPost();
 
+        try {
             $msDataValidator = new SolutionValidation();
 
-            if (! $msDataValidator->validateInputs($msData)) {
+            if (! $msDataValidator->validateInputs($msUpdatedData)) {
                 throw new InvalidInputException($msDataValidator->getErrors());
             }
```
Mauriciocr207, line 114: Validate files here, too

Mauriciocr207 (changes requested):

app/Controllers/CtrlSolution.php
```diff
@@ -20,114 +39,176 @@ public function viewSolutions()
...
-        $measurementSolutionModel = new MeasurementSolutionModel();
-        $measurementSolution      = $measurementSolutionModel->find($id);
+        $msData = (new MeasurementSolutionModel())->getMeasurementSolutionDataWithFiles($msId);
+
+        if (! session()->has('clientData')) {
+            $dataFiles['msIcon']  = [$msData['msIcon']];
+            $dataFiles['msImage'] = [$msData['msImage']];
+        } else {
+            $dataFiles = session()->get('clientData');
+        }
```
LUGAMAFE, line 58: Revert the if always use the true path first
```suggestion
        if (session()->has('clientData')) {
            $dataFiles = session()->get('clientData');
        } else {
            $dataFiles['msIcon']  = [$msData['msIcon']];
            $dataFiles['msImage'] = [$msData['msImage']];
        }
```

app/Controllers/CtrlSolution.php
```diff
@@ -2,14 +2,33 @@
...
+
+    public function __construct()
+    {
+        $fileConfigBuilder = new FileValidationConfigBuilder(self::MEASUREMENT_SOLUTIONS_BASE_ROUTE);
+
+        $fileConfigBuilder->builder('msIcon')->isSVG()->minFiles(1)->maxFiles(1)->maxSize(1, 'MB')->build();
+        $fileConfigBuilder->builder('msImage')->isImage()->minFiles(1)->maxFiles(1)->maxDims(500, 500)->maxSize(2, 'MB')->build();
+
+        $this->fileConfig = $fileConfigBuilder->getConfig();
+    }
```
Mauriciocr207, line 30: To improve and make more dinamic the routes you defined here, codeigniter provides global functions like [url_to](https://www.codeigniter.com/user_guide/helpers/url_helper.html?highlight=url_to#url_to) that allows you pass the class and any of its methods to return the route asociated to it. For example here you can change the controller to something like this:
```suggestion
    private static $MEASUREMENT_SOLUTIONS_BASE_ROUTE;
    private static $MEASUREMENT_SOLUTIONS_CREATE_ROUTE;

    protected FileValidationConfig $fileConfig;

    public function __construct()
    {
        self::$MEASUREMENT_SOLUTIONS_BASE_ROUTE = url_to(self::class . '::viewSolutions');
        self::$MEASUREMENT_SOLUTIONS_CREATE_ROUTE = url_to(self::class . '::viewSolutionCreate');
    
        $fileConfigBuilder = new FileValidationConfigBuilder(self::$MEASUREMENT_SOLUTIONS_BASE_ROUTE);

        $fileConfigBuilder->builder('msIcon')->isSVG()->minFiles(1)->maxFiles(1)->maxSize(1, 'MB')->build();
        $fileConfigBuilder->builder('msImage')->isImage()->minFiles(1)->maxFiles(1)->maxDims(500, 500)->maxSize(2, 'MB')->build();

        $this->fileConfig = $fileConfigBuilder->getConfig();
    }
```

app/Controllers/CtrlSolution.php
```diff
@@ -20,114 +39,176 @@ public function viewSolutions()
...
                 'type'    => 'success',
             ];
 
-            return redirect()->to('/admin/soluciones')->with('response', $response);
+            return redirect()->to(self::MEASUREMENT_SOLUTIONS_BASE_ROUTE)->with('response', $response);
         } catch (InvalidInputException $th) {
-            return redirect()->back()->withInput()->with('errors', $th->getErrors());
+            session()->setFlashdata('clientData', $msUpdatedData);
+
+            return redirect()->to(self::MEASUREMENT_SOLUTIONS_EDIT_ROUTE . $msId)->withInput()->with('errors', $th->getErrors());
```
Mauriciocr207, line 140: Here, in order you use only once the `MEASUREMENT_SOLUTIONS_BASE_ROUTE` variable, then you can do here
```suggestion
            return redirect()->to(url_to(self::class . '::viewSolutionEdit', $msId))->withInput()->with('errors', $th->getErrors());
```

Mauriciocr207 (changes requested):
I left you some comments
CC: @LUGAMAFE

LUGAMAFE (approved):
Just one small nitpick and same comments as @Mauriciocr207 before doing merge

Cristianpan (approved):

CarlosMay7 (approved):

app/Controllers/CtrlSolution.php
CarlosJCD: why not using the static constant generated above?

app/Controllers/CtrlSolution.php
CarlosJCD: can I generate its value on the variable definitions instead of the constructor? or is it really necessary to generate it on the class constructor?

Mauriciocr207: The correct way to pass the method to the url_to is addin a string `"Controller::method"` , when we use `self::class`, that generates a string with the controller class name

app/Controllers/CtrlSolution.php
Mauriciocr207: When we use constants like the way you do before, we can´t use functions in constants expresions
Like: 
```php
private const MS_BASE = url_to(...)
```  
That is incorrect because is not a constant argument, but a function. That´s the reason we need to set an static variable and set it into the `__construct()` function

Mauriciocr207 (approved):

PR #175, opened 2024-01-03
KIN-305 delete certificate functionality certificates page

FernandoJoachin:
[KIN-305](https://loopcrack.atlassian.net/browse/KIN-305) Small [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

As a company I would like to delete certificates from the system so that I can remove any certificate that I no longer have.

### Solution

Add remove certificate functionality in CtrlCertificate

### Requirements checklist

- [x] It must remove the certificate record from the database
- [x] It must return a response with a message indicating success on the deleting the certificate or failure in the process
- [x] It must delete the certificates files from the server
- [x] It must remove the certificate preview image and the certificate file record from the database

### Testing

- [x] Go to the next route /admin/certificados/crear and create a certificate. After creating the certificate, delete it with the corresponding button
- [x] Check that the files were deleted on both the server and the database

[KIN-305]: https://loopcrack.atlassian.net/browse/KIN-305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Mauriciocr207 (approved):

CarlosMay7 (approved):

app/Controllers/CtrlCertificate.php
```diff
@@ -166,17 +167,29 @@ public function updateCertificate(string $certificateId)
...
+            $certificate      = $certificateModel->getCertificate($certificateId);
+
+            if (empty($certificate)) {
+                throw new Exception();
+            }
+
+            $certificateModel->deleteCertificate($certificate);
+
+            FileManager::deleteMultipleFoldersWithContent([$certificate['previewUuid']]);
+            FileManager::deleteMultipleFoldersWithContent([$certificate['certificateUuid']]);
```
CheloChidoone, line 182: I think it is better if you just call this function once and pass both uuids in the array you give it, calling it twice should not be necessary

CheloChidoone (changes requested):
@FernandoJoachin I left a small comment for you to check and also make sure to check the other observation made by Cristian

app/Models/CertificateModel.php
```diff
@@ -68,6 +69,32 @@ public function updateCertificate(string $certificateId, array $certificateData)
...
     }
 
+    public function deleteCertificate(array $certificate)
+    {
+        try {
+            $this->db->transException(true)->transStart();
+            $isDeleted = $this->delete($certificate['certificateId']);
+            if (! $isDeleted) {
+                throw new Exception();
+            }
```
Cristianpan, line 79: It's unnecessary to verify that the element was deleted and then throw an exception because if it isn't deleted, the statement already throws an exception.

app/Models/CertificateModel.php
```diff
@@ -68,6 +69,32 @@ public function updateCertificate(string $certificateId, array $certificateData)
...
+
+            $fileModel = new FileModel();
+
+            $queryWhere = "uuid={$this->escape($certificate['previewUuid'])} OR uuid={$this->escape($certificate['certificateUuid'])}";
+            $isDeleted  = $fileModel->where($queryWhere)->delete();
```
Cristianpan, line 84: Instead of creating an or manually, you can make use of the whereIn statement
```suggestion
            $isDeleted  = $fileModel->whereIn('uuid', [$certificate['previewUuid'], $$certificate['certificateUuid']])->delete();
```

Cristianpan (changes requested):
I left you some comments, please check it and make the changes

LUGAMAFE (approved):

CheloChidoone (approved):
Good Job @FernandoJoachin

Cristianpan (approved):

PR #204, opened 2024-01-22
KIN-400 Fix Meta Image Generator

CheloChidoone:
[KIN-400](https://loopcrack.atlassian.net/browse/KIN-400) Small [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

The Meta Image Generator does not work properly for categories and may raise a warning for some images like the new category jumbotron image

### Solution

Fix the Meta Image Generator integration for the catalog pages and prevent warnings during the process

### Requirements checklist

- [ ] Meta Image generates correctly when a main category is selected in the catalog
- [ ] Meta Image generator does not show warnings when generating

### Testing

To test you need to have a category with properly uploaded images for it
- [ ] Select the category for your test from the `equipos` page, check that on the same folder where the jumbotron image for the category was created the new meta image was generated successfully

## Resources

None

## Demo

None

[KIN-400]: https://loopcrack.atlassian.net/browse/KIN-400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/Utils/OgImageGenerator.php
```diff
@@ -2,6 +2,8 @@
 
 namespace App\Utils;
 
+use Exception;
```
CarlosJCD, line 5: please remove this unused import
```suggestion
```

CarlosJCD (changes requested):
For some reason the jumbotron when there is no filter selected does not load. I am not sure if this is an error of my local environment so please let me know if it is.

Screenshot 2024-01-23 at 23 19 22

app/Utils/OgImageGenerator.php
CheloChidoone: I removed the import @CarlosJCD thanks for the observation

CheloChidoone: About the jumbotron issue I hadn't seen it because it worked fine on the server but now I noticed that when there is no category filter selected the jumbotron tries to load the image from this path and we don't have this image on this path on our locals so we see the issue, `/uploads/4972868bb3df514edd5939b872e1d7a5/Bannersproductos-Recuperado.psdMesa-de-trabajo-1.png`, why did you choose to do it this way @Mauriciocr207 ??

Cristianpan: I think the error is because the image wasn't uploaded correctly in uploads when the filter ticket was uploaded. Please confirm if this is the case or not @Mauriciocr207, @CarlosMay7 , @Oswaldo-Chan

Mauriciocr207: @LUGAMAFE told me that I get that image to show the jumbotron image in case there is not filter applied. In order the image is loaded now on server, I just choose the path on production, because tha image was ready loaded there.

Mauriciocr207: I think that image could be into the image folder instead of uploads folder.

Mauriciocr207 (approved):

CarlosMay7 (approved):

app/Utils/OgImageGenerator.php
FernandoJoachin: And was what Carlos Calderón mentioned solved? Because what he mentioned keeps happening to me

CheloChidoone: @FernandoJoachin we don't have that image in our locals because they took the path from the uploads on the production server, I'll add it to the image folder like Mau said instead of uploads and that should fix it

Oswaldo-Chan (approved):

CarlosJCD (approved):

app/Controllers/CtrlPublicPages.php
```diff
@@ -152,19 +152,26 @@ public function viewCategory(): string
                 'selected' => $perPageOption === $filter['por-pagina'],
             ];
         }, $this->allowedPerPageProducts);
-          
-        $metaImage = OgImageGenerator::generateOgImage($category['fileRoute']);
+
+        if($principalCategory){
+            $metaImage = OgImageGenerator::generateOgImage($principalCategory['fileRoute']);
+            $metaDescription = 'Explora los mejores productos de '.$principalCategory['categoryName'];
```
LUGAMAFE, line 158:
```suggestion
            $metaDescription = "Explora los mejores productos de {$principalCategory['categoryName']}";
```

LUGAMAFE (approved):
Nice

PR #205, opened 2024-01-23
KIN-405 add date column on emails table of the admin panel

FernandoJoachin:
[KIN-405](https://loopcrack.atlassian.net/browse/KIN-405) Small [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=loopcrack-org&utm_campaign=badge_info)

### Problem

Data for emails sent from the contact form, support form, and product form should be displayed with their sent date.

### Solution

Display the sent date of the emails 

### Requirements checklist

- [x] The sent date must be displayed on the emails table.

### Testing

- [x] Complete a form and then navigate to admin/emails. There, the system should display the sent email along with its sending date.

[KIN-405]: https://loopcrack.atlassian.net/browse/KIN-405?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CarlosJCD (approved):

app/Views/admin/emails/Emails.php
```diff
@@ -61,11 +62,16 @@
...
                                         <td>
-                                            <div class="text-<?= ($email['emailTypeId'] === '1') ? 'warning' : (($email['emailTypeId'] === '2') ? 'info' : 'danger'); ?> text-wrap">
+                                            <div class="text-<?= ($email['emailTypeId'] === '1') ? 'warning-emphasis' : (($email['emailTypeId'] === '2') ? 'info-emphasis' : 'success-emphasis'); ?> text-wrap">
                                                 <?= ($email['emailTypeId'] === '1') ? 'Contacto' : (($email['emailTypeId'] === '2') ? 'Soporte técnico' : 'Información producto'); ?>
                                             </div>
 
                                         </td>
+                                        <td>
+                                            <div class="d-flex align-items-center fw-medium">
+                                                <p class="text-wrap"><?= $email['createdAt']; ?></p>
```
Cristianpan, line 72: I think you should format the date better as "day/month/year"
```suggestion
  <p class="text-wrap"><?= date('j/m/Y', strtotime($email['createdAt']))?></p>
```

app/Views/admin/emails/Emails.php
```diff
@@ -61,11 +62,16 @@
                                             </div>
                                         </td>
                                         <td>
-                                            <div class="text-<?= ($email['emailTypeId'] === '1') ? 'warning' : (($email['emailTypeId'] === '2') ? 'info' : 'danger'); ?> text-wrap">
+                                            <div class="text-<?= ($email['emailTypeId'] === '1') ? 'warning-emphasis' : (($email['emailTypeId'] === '2') ? 'info-emphasis' : 'success-emphasis'); ?> text-wrap">
```
Cristianpan, line 65: I see that you are using type identifiers and I think you should replace those magic numbers with constant values to improve the readability of the code. I remember that in the controller where we send the emails, we have those constants but they are of the class. Since they would now be used in more than one place, maybe it would be good to move them to App/Config/Constant.php. But I don't know what @LUGAMAFE , @CheloChidoone @Mauriciocr207, @CarlosJCD think

Cristianpan (changes requested):

app/Views/admin/emails/Emails.php
Mauriciocr207: Maybe for this, as you have the hour, you can set the date on that format and the hour like this:
```suggestion
<p class="text-wrap"><?= date('j/m/Y, g:i a', strtotime($email['createdAt']))?></p>
```

In case you want more date format recomendations, you have thie reference @FernandoJoachin: [date on php](https://www.php.net/manual/es/function.date.php)

app/Views/admin/emails/Emails.php
CarlosJCD: I missed that out.! I agree, it is important to avoid magic numbers/strings.

Mauriciocr207: I agree, but also think we could remove the if-else operation with the ternary operator and change by
```php
<?php foreach ($emails as $email) { ?>
<?php
switch ($email['emailTypeId']) {
    case CONTACT_EMAIL_ID:
        $emailTypeEmphasis = 'text-warning-emphasis';
        $emailTypeText     = 'Contacto';
        break;

    case SUPPORT_EMAIL_ID:
        $emailTypeEmphasis = 'text-info-emphasis';
        $emailTypeText     = 'Soporte técnico';
        break;

    case PRODUCT_EMAIL_ID:
        $emailTypeEmphasis = 'text-success-emphasis';
        $emailTypeText     = 'Información de producto';

        break;
}
?>
<tr>
    <!-- Rest of code -->
    ...
    <td>
        <div class="<?= $emailTypeEmphasis?> text-wrap">
            <?= $emailTypeText?>
        </div>
    </td>
```
Or something similar. Also in this way you don´t need to evaluate the emailTypeId twice, just once.

Cristianpan: Hmmm maybe would be better to create an associative array instead of a switch but I think that the text of the email type should be taken directly from the database that is already being stored there.

CheloChidoone: @Mauriciocr207 @Cristianpan I think the if-else operation is fine like we have it but you do need to avoid the magic numbers @FernandoJoachin, if we used the email type id numbers as constants in multiple places then we can move those constants to `App/Config/Constant.php` like Cristian said and used them from there, make those changes and this should be good.

CheloChidoone (approved):
Good Job @FernandoJoachin, just finish the change Cristian requested regarding the magic numbers

Mauriciocr207 (approved):
Good job! Just make the requested changes

Cristianpan:
Thanks for changing the magic numbers to constants, but I think you should also add those constants in `app/Controllers/CtrlEmail` and remove the class constants. I know it doesn't pertain to the ticket but I think it would be nice if you did that since it's a small change.

app/Views/admin/emails/Emails.php
```diff
@@ -61,11 +62,16 @@
                                             </div>
                                         </td>
                                         <td>
-                                            <div class="text-<?= ($email['emailTypeId'] === '1') ? 'warning' : (($email['emailTypeId'] === '2') ? 'info' : 'danger'); ?> text-wrap">
-                                                <?= ($email['emailTypeId'] === '1') ? 'Contacto' : (($email['emailTypeId'] === '2') ? 'Soporte técnico' : 'Información producto'); ?>
+                                            <div class="text-<?= ($email['emailTypeId'] === EMAIL_CONTACT_FORM) ? 'warning-emphasis' : (($email['emailTypeId'] === EMAIL_TECHNICAL_SUPPORT_FORM) ? 'info-emphasis' : 'success-emphasis'); ?> text-wrap">
+                                                <?= ($email['emailTypeId'] === EMAIL_CONTACT_FORM) ? 'Contacto' : (($email['emailTypeId'] === EMAIL_TECHNICAL_SUPPORT_FORM) ? 'Soporte técnico' : 'Información producto'); ?>
                                             </div>
```
LUGAMAFE, line 67:
```suggestion
                                            <?php
                                                switch ($email['emailTypeId']) {
                                                    case EMAIL_CONTACT_FORM:
                                                        $class = 'warning-emphasis';
                                                        $text = 'Contacto';
                                                        break;
                                                    case EMAIL_TECHNICAL_SUPPORT_FORM:
                                                        $class = 'info-emphasis';
                                                        $text = 'Soporte técnico';
                                                        break;
                                                    default:
                                                        $class = 'success-emphasis';
                                                        $text = 'Información producto';
                                                        break;
                                                }
                                            ?>

                                            <div class="text-<?= $class; ?> text-wrap">
                                                <?= $text; ?>
                                            </div>
```

LUGAMAFE (approved):
Good Job @FernandoJoachin

CarlosMay7 (approved):

LUGAMAFE:
@Cristianpan can you check this PR please

Cristianpan (approved):

PR #207, opened 2024-01-23
KIN-404 improve home page layout

CarlosMay7:
### Problem

The home page had some design details that need to be resolved.

### Solution

Changing some logic and styles on the home page 

### Requirements checklist

- [ ] When no measurement solutions exist, the section header must disappear.
- [ ] The error messages of the contact form no longer misalign the inputs of the form
- [ ] The popup message when an email is sent is more descriptive.
- [ ] The size of the popup message when an email is sent is decreased.

### Testing

- [ ] Manually check the form error messages
- [ ] Manually check the popup message when the contact email is sent

## Demo
[Demo Home Page](https://drive.google.com/file/d/13IDFI2JGOR2F81iNj3riZq1145chYmrL/view?usp=sharing)


CarlosJCD (approved):
Nice changes

CheloChidoone (changes requested):
Good Job @CarlosMay7, I just think the message could be more complete, try adding something like what I suggested, it doesn't have to be exactly like my suggestion but something that communicates to the user that we will respond to their contact request soon.

Mauriciocr207 (commented):
I see that if you have less than four measurement solutions, then the cards goes to left
image

I think we can add a `justify-content:center` css property but only when we have less than four cards. In other case the cards doesn´t position correctly
image

app/Controllers/CtrlEmail.php
```diff
@@ -52,7 +52,7 @@ public function sendContactEmail(): RedirectResponse
         if ($successEmail) {
             $response = [
                 'title'   => 'Envío exitoso',
-                'message' => 'Se ha enviado correctamente',
+                'message' => 'Gracias por ponerse en contacto con nosotros',
```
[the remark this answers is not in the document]
LUGAMAFE: 'El Formulario se ha enviado a nuestro equipo, trataremos de ponernos en contacto con usted lo más pronto posible.'

src/public/sass/ITCSS/06_components/pages/home/kinub-video/_kinub-video.scss
```diff
@@ -15,29 +15,26 @@
...
     }
   }
   @include desktopXXL() {
-    margin: 10rem 0;
     &__container {
-      max-width: 60vw;
+      max-width: 45vw;
     }
   }
 }
```
LUGAMAFE, line 40: Video should use the full viewport, what was needed here is to verify laptop's views because the video was cutting off.

LUGAMAFE (changes requested):

src/public/sass/ITCSS/06_components/pages/home/kinub-video/_kinub-video.scss
```diff
@@ -4,40 +4,23 @@
...
-  @include desktopXL() {
-    margin: 10rem 0;
-    scroll-margin-top: 4rem;
-    &__container {
-      max-width: 75vw;
-    }
-  }
-  @include desktopXXL() {
-    margin: 10rem 0;
+    margin: 6rem 0;
```
LUGAMAFE, line 18: you removed the scroll snaping @CarlosMay7 :( 
scroll-snap-align
and 
scroll-margin-top

LUGAMAFE (changes requested):

app/Views/public/index.php
```diff
@@ -38,7 +38,7 @@
     </div>
 </section>
 
-<section class="measurement-solutions">
+<section class="measurement-solutions <?= empty($measurementSolutions) ? 'measurement-solutions--empty' : ''; ?>">
```
Cristianpan, line 41:
```suggestion
<?php if (empty($measurementSolutions)): ?>
<section class="measurement-solutions">
<!-- html -->
</section>
<?php endif ?>
```
Instead of making a condition to put a class with `display:none` attribute you could condition to paint or not that section in the view.

CheloChidoone (changes requested):
@CarlosMay7 Good Job in General but I think when you made the change Mau requested you forgot to think about the mobile styles and when switching to mobile it now looks like this, you should fix it so it looks like we had it before.
image

Cristianpan (changes requested):

src/public/sass/ITCSS/06_components/pages/home/measurement-solutions/_swiper.scss
```diff
@@ -4,6 +4,11 @@
   width: 100%;
   height: 36vh;
 }
+
+.swiper-wrapper--center {
+  justify-content: center;
+}
+
```
Mauriciocr207, line 11:
```suggestion
.swiper-wrapper--center {
  @include phoneXL() {
    justify-content: center;
  }
}
```
This could solve the issue @CheloChidoone say.

src/public/js/home.js
```diff
@@ -29,6 +29,17 @@ new Swiper('.swiper', {
...
+  on: {
+    init: () => {
+      const swiperWrapper = document.querySelector('.swiper-wrapper');
+      const slidesCount = swiperWrapper.children.length;
+
+      if (slidesCount < 4) {
+        swiperWrapper.classList.add('swiper-wrapper--center');
+      }
+    },
+  },
```
Mauriciocr207, line 42: This solution could be done with php as well, counting the records that are received and put the necessary class if there are less than 4 records. I think it is a better solution in order you can count the real records on db and not count the elements on html. What do you think?

Mauriciocr207 (approved):
Good job!, just make the requested changes.

Cristianpan (approved):

CheloChidoone (approved):

Oswaldo-Chan (approved):

CarlosJCD:
@LUGAMAFE Can you help us approve these please?

FernandoJoachin (approved):

CarlosJCD (approved):

PR #218, opened 2024-01-29
KIN-413 Improve 404 page

FernandoJoachin:
### Problem

The actual 404 page does not match with the current design of the public pages.

### Solution

Implement a better 404 page that matches with the current design of the public pages.

### Requirements checklist

- [x] The 404 page must contain the public view header.
- [x] The 404 page must contain the public view footer.
- [x] The 404 page must contain the whatsapp button.

### Testing

- [x] Access a url that does not exist

## Demo
[](https://postimg.cc/vgKgzGPK)


LUGAMAFE (changes requested):
The page is so basic and the message too long and complex, no one is going to read this page.

image

Cristianpan:
@FernandoJoachin, you can find some images you can use for 404 page in the following link: https://www.freepik.es/search?format=search&last_filter=query&last_value=404&query=404

Cristianpan (commented):
Nice job! It looks better than last time but I have one observation. On some mobile devices like the !phone 12 pro, it doesn't take the full viewport height, which gives a bad look. 
image

Likewise, you should perform image optimization with gulp. To do this add the image first in `src/image` and then do `npm run images`, which will generate the optimized images in `assets/images`.

To use the optimized images you should do some like this: 
```html 
    <picture>
         <source srcset="example.avif" type="image/avif">
         <source srcset="example.webp" type="image/webp">
         <img loading="lazy" src="example.png" >
    </picture>
```

Cristianpan (changes requested):
Check the comments I've left you please

src/public/sass/ITCSS/06_components/pages/error-404/_error-404.scss
```diff
@@ -0,0 +1,46 @@
...
+@use '../../../01_settings/fonts' as *;
+
+.error-404 {
+  max-width: 50rem;
+  padding: 0 1rem;
+  margin: 3rem auto;
+  color: $color-gray;
+  font-family: $brand-font-tertiary;
+  font-weight: $weight-light;
+  text-align: center;
```
Mauriciocr207, line 12: Here, when Oswaldo fix the problem with the fixed footer, then you could add some styles in order the box of `.error-404` be centered like this:
image

Also, wee see that fonts are not the same as the rest of this page. Please, remove the font-family change. 
```suggestion
  max-width: 50rem;
  padding: 0 1rem;
  margin: 3rem auto;
  color: $color-gray;
  font-weight: $weight-light;
  text-align: center;
```

On this suggestion, I changed the styles of footer to fixed to bottom. But actually, you don´t need to modify footer styles because Oswaldo fixed this on other pull request.

Mauriciocr207 (changes requested):
I left you some comments, good job! 🤙
CC: @CarlosMay7

app/Views/errors/html/error_404.php
```diff
@@ -1,84 +1,25 @@
...
+<?php $this->endSection(); ?>
+
+<?php $this->section('content'); ?>
+    <main class="error-404">
+        <picture>
+            <source srcset="/assets/images/404.avif" type="image/avif">
+            <source srcset="/assets/images/404.webp" type="image/webp">
+            <img loading="lazy" class="error-404__image" src="/assets/images/404.jpg" alt="Error 404">
+        </picture>
+        <h1 class="error-404__heading">Error 404: Página no encontrada</h1>
```
CheloChidoone, line 19:
```suggestion
        <h1 class="error-404__heading">Página no encontrada</h1>
```

CheloChidoone (approved):
Good Job @FernandoJoachin, I suggest a small change in the Heading since the image already says 404 error I think it is not necessary to repeat that. Make the change before merging, otherwise it looks much better now.

src/public/sass/ITCSS/06_components/pages/error-404/_error-404.scss
```diff
@@ -0,0 +1,49 @@
...
+
+  &__message {
+    margin-top: 1rem;
+  }
+
+  &__btn {
+    padding: 0.375rem 0.625rem;
+    border: none;
+    border-radius: 0.3125rem;
+    margin-top: 1.5rem;
```
CarlosMay7, line 37: Maybe you can reduce this value
```suggestion
    margin-top: 0.5rem;
```

CarlosMay7 (approved):
Good job, just check some suggestions I made you

src/public/sass/ITCSS/06_components/pages/error-404/_error-404.scss
```diff
@@ -0,0 +1,49 @@
...
+    margin-bottom: 0;
+    color: $color-black;
+    font-size: $font-size-h1;
+  }
+
+  &__message {
+    margin-top: 1rem;
+  }
+
+  &__btn {
```
CarlosMay7, line 33: Just a suggestion, maybe you can change the font size to not have a bigger button on the Galaxy fold

image

CarlosJCD (approved):

Cristianpan (approved):
Good job! Just check out the comments May left you.

